PR #12, opened 2025-08-13
Fix overly verbose version constraint

jlaswell:
`^8.1` is equivalent to `^8.1|^8.2`


jlaswell:
It seems like `main` is broken?

PR #27, opened 2025-09-03
Clarify MCP Inspector installation instructions

nunomaduro:


nunomaduro:
you are correct:

```
npm warn exec The following package was not found and will be installed: @modelcontextprotocol/inspector@0.16.6
```

PR #71, opened 2025-09-30
Implement validated method in Request.php

nicobleiler:
Adds a method `validated` similar to formRequest so you can simply define the rules, messages and attributes & call validated on the request instead of having to rewrite it.

Example MCP Request
```php
<?php

namespace App\Mcp\Requests;

use Laravel\Mcp\Request;

class QueryMCPRequest extends Request
{
    public function rules()
    {
        return [
            'query' => 'required|string',
            'page' => 'nullable|page',
            'pageSize' => 'nullable|page_size',
        ];
    }
}
```

Example handling in a tool
```php
    ...
    public function handle(QueryMCPRequest $request): Response
    {
        $validated = $request->validated();
    ...
```

### DIFF
--- a/src/Request.php
+++ b/src/Request.php
@@ -76,6 +76,48 @@ class Request implements Arrayable
         return Validator::validate($this->all(), $rules, $messages, $attributes);
     }
 
+    /**
+     * Get custom rules for validator errors.
+     *
+     * @return array
+     */
+    public function rules()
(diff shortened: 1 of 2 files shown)


taylorotwell:
We would need to think through this more thoroughly regarding auto-validation like form requests, etc.

PR #76, opened 2025-10-07
Add Macroable and Conditionable traits to some core classes

mattwells-coex:
- Applied `Macroable` and `Conditionable` traits to:
  - `Request`
  - `Response`
  - `TestResponse`
- Verified support via new tests for:
  - Custom macro methods (`->foo()`)
  - `when()` and `unless()` conditional method chaining


taylorotwell:
You don't need these tests. It's testing things that are already tested by the framework.

PR #87, opened 2025-10-16
Feature: adds security to the OAuth registration endpoint

jsandfordhughescoop:
This PR enhances the OAuth client registration endpoint by adding validation and configurable redirect domain restrictions.

### Changes
- Validate redirect URIs against configured allowed domains.
- Added `config/mcp.php` with:
  - `allow_all_redirect_domains` — toggle to allow all redirect domains.
  - `allowed_redirect_domains` — list of whitelisted domains.

### Benefits
- **Improved security:** prevents registration with unauthorized redirect URIs.  
  *This is effectively the only meaningful security enforcement available within the MCP specification, which otherwise does not provide mechanisms for authenticating or authorizing clients during registration.*

### Notes
I have not validated client name as this could be a breaking change to users who have overwritten the default of Passport functionality.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -80,7 +80,7 @@
             "pint --test",
             "rector --dry-run"
         ],
-        "test:unit": "pest --ci --coverage --min=91.5",
+        "test:unit": "pest --ci --coverage --min=91.2",
         "test:types": "phpstan",
         "test": [
             "@test:lint",


barryvdh:
I think the request parameter client_name got accidentally changed to name
Before:
https://github.com/laravel/mcp/pull/87/files#diff-92678f35c558701ce8a360588eb08f99026231c8b32cd0d118315b84ea75efb0L113
After: 
https://github.com/laravel/mcp/pull/87/files#diff-728478d0055c4873d77a0804b2b3c8663a55a6204c39ea99063a57ab91573f85R40

The tests pass because not typehints are on the fake registry, but in practice they fail. 

My colleague has added a fix in https://github.com/laravel/mcp/pull/104

PR #103, opened 2025-11-06
Add App response type

zacksmash:
A new Response method has been added called `app()`, to be used in Resource primitives for returning the required data to a ChatGPT app. An Enum has been added as well to assist with setting common OpenAI Apps SDK related values.

Example:

```php
class WeatherAppResource extends Resource
{
    const TEMPLATE = 'ui://apps/weather';
    
    protected string $uri = self::TEMPLATE;

    protected string $mimeType = OpenAI::MIME_TYPE->value;

    // ...

    /**
     * Handle the resource request.
     */
    public function handle(): Response
    {
        return Response::app(view('mcp.app'),
            fn (App $app) => $app->prefersBorder()
        );
    }
```

>[!TIP]
> `$app` includes some helpers for ChatGPT apps, but also includes  a `meta()` method to allow adding additional items to the metadata response

This relies on #102, as you'll need to use the `meta()` and `structuredContent()` methods, on the Tool Response. `meta()` is used to load data into your app that is not seen by the Model/LLM. `structuredContent()` is used to hydrate your app with data, and provide the Model/LLM with contextual information about the state of the app. [Docs Reference](https://developers.openai.com/apps-sdk/build/mcp-server#structure-the-data-your-tool-returns)

```php
#[IsReadOnly()]
class WeatherTool extends Tool
{

    ...

    /**
     * Handle the tool request.
     */
    public function handle(Request $request, WeatherData $weatherData): Response|array
    {
        $city = $request->get('city', 'San Francisco');

        $content = $weatherData->handle($request);

        return [
            Response::json($content), // Required for backwards compatibility with MCP
            Response::text("Here is the current weather information you requested for {$city}.")
                ->meta(['route' => 'weather'])
                ->structuredContent($content),
        ];
    }
}
```

This also relies on #101, as you'll need to use the `$meta` property in the Tool Definition to add additional required data for [ChatGPT Apps](https://developers.openai.com/apps-sdk/build/mcp-server#point-to-a-component-template):

```php
#[IsReadOnly()]
class WeatherTool extends Tool
{

    ...

    /**
     * Get the tool's meta information.
     *
     * @return array<string, mixed>
     */
    public function meta(): array
    {
        return [
            OpenAI::OUTPUT_TEMPLATE->value => WeatherAppResource::TEMPLATE,
            OpenAI::TOOL_INVOKING->value => 'Working on it...',
            OpenAI::TOOL_INVOKED->value => 'Weather Tool Completed',
        ];
    }
}
```

Ref #99

### DIFF
--- a/tests/Unit/Content/AppTest.php
+++ b/tests/Unit/Content/AppTest.php
@@ -2,7 +2,6 @@
 
 use Laravel\Mcp\Enums\OpenAI;
 use Laravel\Mcp\Server\Content\App;
-use Laravel\Mcp\Server\Content\Text;
 use Laravel\Mcp\Server\Prompt;
 use Laravel\Mcp\Server\Resource;
 use Laravel\Mcp\Server\Tool;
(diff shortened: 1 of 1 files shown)


zacksmash:
There might be a better way to handle this, for ChatGPT Apps and MCP-UI. I'll tinker and come back to this.

PR #104, opened 2025-11-07
Fix client_name rename in oauth registrar

mikebouwmans:
This PR fixes the breaking refactor (PR #87)that renamed the field `client_name` to `name` which is incorrect following the spec: https://datatracker.ietf.org/doc/html/rfc7591#section-3.1

We're still falling back to 'name' to not make this a breaking change. Not sure if that is what we want to support.

Also fixed the tests since it wasn't strict enough without the typehints of `createAuthorizationCodeGrantClient`.

### DIFF
--- a/tests/Unit/Server/RegistrarTest.php
+++ b/tests/Unit/Server/RegistrarTest.php
@@ -166,7 +166,7 @@ it('handles oauth registration endpoint', function (): void {
         eval('
             namespace Laravel\Passport;
             class ClientRepository {
-                public function createAuthorizationCodeGrantClient($name, $redirectUris, $confidential, $user, $enableDeviceFlow) {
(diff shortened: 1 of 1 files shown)


src/Server/Http/Controllers/OAuthRegisterController.php
```diff
@@ -37,7 +37,7 @@ public function __invoke(Request $request): JsonResponse
         );
 
         $client = $clients->createAuthorizationCodeGrantClient(
-            name: $request->get('name'),
+            name: $request->get('client_name', $request->get('name')),
```
barryvdh: imho we can remove the fallback, because it's not according to the spec. But we can check what Taylor prefers

pushpak1300: @taylorotwell @joetannenbaum what do you think about this?
I’m leaning toward removing the fallback, but that would be a breaking change unless we’re okay introducing it now or want to handle it with a fallback in the interim.

PR #105, opened 2025-11-10
fix: allow multi-segment issuer paths

isaac-bowen:
### What does this PR do?
Adds support for multi-segment issuer paths in the OAuth well-known endpoints registered by `Registrar::oauthRoutes()`.

### Why is this needed?
According to the OAuth 2.1 / OpenID Connect Discovery specification, the `issuer` URI may contain path segments (e.g. `https://api.example.com/mcp/weather`).  
Previously, the route definition only matched a single path segment, so requests to nested issuers like `/mcp/weather` were not discoverable.

This patch updates both the `/.well-known/oauth-protected-resource/{path?}` and `/.well-known/oauth-authorization-server/{path?}` routes to allow multi-segment paths.

### How does this benefit users?
- Enables mounting MCP OAuth discovery routes under nested prefixes (e.g. `https://example.com/mcp/weather`).  
- Maintains backward compatibility for single-segment and root-level issuers.  
- Aligns with OAuth 2.1 / OIDC Discovery, which permits issuer URIs with sub-paths.  

### What changed?
- Updated the `.well-known` routes to use `.where('path', '.*')`, allowing multiple path segments.  
- No behavioural changes for existing single-segment or root-level issuers.

### Breaking changes?
None.

### Tests
Added new tests in `tests/Unit/Server/RegistrarTest.php`:
- `it('handles oauth discovery with multi-segment paths')`
- `it('handles oauth discovery with single segment paths')`
- `it('handles oauth discovery with no path')`

These verify correct JSON responses and ensure backward compatibility across all cases.

### References
- [OpenID Connect Discovery 1.0 §2](https://openid.net/specs/openid-connect-discovery-1_0-final.html#:~:text=optionally%2C%20port%20and%20path%20components) — issuer URIs may include path components.  
- [Laravel MCP documentation](https://laravel.com/docs/12.x/mcp#web-servers:~:text=Mcp%3A%3Aweb(%27/-,mcp/weather,-%27%2C%20WeatherServer%3A%3Aclass)%3B) — official example shows `/mcp/weather` as a nested path.

### DIFF
--- a/src/Server/Registrar.php
+++ b/src/Server/Registrar.php
@@ -89,7 +89,7 @@ class Registrar
             'resource' => url('/'.$path),
             'authorization_servers' => [url('/'.$path)],
             'scopes_supported' => ['mcp:use'],
-        ]))->name('mcp.oauth.protected-resource');
+        ]))->where('path', '.*')->name('mcp.oauth.protected-resource');
 
         Router::get('/.well-known/oauth-authorization-server/{path?}', fn (?string $path = '') => response()->json([
             'issuer' => url('/'.$path),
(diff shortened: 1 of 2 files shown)


isaac-bowen:
We ran into this when connecting our MCP to Claude, which only supports auto-discovery. When it tried to find our issuer, `mcp/govtribe`, it was coming back with 404 even though that should be a valid issuer format.

The same issue happens with the weather example from the documentation.

PR #111, opened 2025-11-19
Add Annotation Support on Resources

pushpak1300:
This PR adds support for resource annotations as defined in the [MCP Specs](https://modelcontextprotocol.io/specification/2025-06-18/server/resources#annotations).

## Usage

Three annotation types are now available for use with Resource classes.

### 1. Audience.

```php
use Laravel\Mcp\Enums\Role;
use Laravel\Mcp\Server\Annotations\Audience;
use Laravel\Mcp\Server\Resource;

#[Audience(Role::ASSISTANT)]
class SystemLogsResource extends Resource
{
    protected string $description = 'System logs for debugging';

    public function handle(): Response
    {
        return Response::text($this->getSystemLogs());
    }
}
```

Accepted values: `Role::USER`, `Role::ASSISTANT`, or an array of both.

```php
#[Audience([Role::USER, Role::ASSISTANT])]
class DocumentationResource extends Resource { }
```

### 2. Priority.

A numerical score between 0.0 and 1.0 indicating resource importance.

```php
use Laravel\Mcp\Server\Annotations\Priority;

#[Priority(1.0)]
class CriticalConfigResource extends Resource
{
    protected string $description = 'Critical application configuration';

    public function handle(): Response
    {
        return Response::text(config('app'));
    }
}
```

Validation. Throws `InvalidArgumentException` if value is outside the 0.0 to 1.0 range.

### 3. LastModified

An ISO 8601 timestamp showing when the resource was last updated.

```php
use Laravel\Mcp\Server\Annotations\LastModified;

#[LastModified('2025-01-12T15:00:58Z')]
class CachedDataResource extends Resource
{
    protected string $description = 'Cached analytics data';

    public function handle(): Response
    {
        return Response::text($this->getCachedData());
    }
}
```

Accepted formats:

- `2025-01-12T15:00:58Z`
- `2025-01-12T15:00:58+00:00`
- Any other ISO 8601 compliant format

Validation. Throws `InvalidArgumentException` if the timestamp format is invalid.

## Combining Multiple Annotations

All three annotations can be combined on a single resource.

```php
use Laravel\Mcp\Enums\Role;
use Laravel\Mcp\Server\Annotations\Audience;
use Laravel\Mcp\Server\Annotations\LastModified;
use Laravel\Mcp\Server\Annotations\Priority;

#[Audience(Role::USER)]
#[Priority(0.9)]
#[LastModified('2025-01-12T15:00:58Z')]
class UserDashboardResource extends Resource
{
    protected string $description = 'User-facing dashboard data';

    public function handle(): Response
    {
        return Response::text($this->getDashboardData());
    }
}
```

## How It Works

Annotations are automatically included in resource listings via the `resources/list` method.

```json
{
  "uri": "file://resources/user-dashboard",
  "name": "UserDashboard",
  "description": "User-facing dashboard data",
  "mimeType": "text/plain",
  "annotations": {
    "audience": ["user"],
    "priority": 0.9,
    "lastModified": "2025-01-12T15:00:58Z"
  }
}
```

If no annotations are present, the `annotations` field is omitted entirely from the response. 

> [!IMPORTANT]
>  This implementation throw error if Invalid Annotations are added on the `Tool::class` or `Resource::class`

### DIFF
--- a/src/Server/Annotations/Annotation.php
+++ b/src/Server/Annotations/Annotation.php
@@ -6,9 +6,6 @@ namespace Laravel\Mcp\Server\Annotations;
 
 use Laravel\Mcp\Server\Contracts\Annotation as AnnotationContract;
 
-/**
- * @property mixed $value
- */
 abstract class Annotation implements AnnotationContract
 {
     //
--- a/src/Server/Annotations/Audience.php
+++ b/src/Server/Annotations/Audience.php
@@ -20,14 +20,17 @@ class Audience extends Annotation
      */
     public function __construct(Role|array $roles)
     {
-        $this->value = array_map(
-            fn ($role) => $role instanceof Role
-                ? $role->value
-                : throw new InvalidArgumentException(
+        $roles = Arr::wrap($roles);
+
+        foreach ($roles as $role) {
+            if (! $role instanceof Role) {
+                throw new InvalidArgumentException(
                     'All values of '.Audience::class.' attributes must be instances of '.Role::class
-                ),
-            Arr::wrap($roles)
-        );
+                );
+            }
+        }
+
+        $this->value = array_map(fn (Role $role) => $role->value, $roles);
     }
 
     public function key(): string
--- a/src/Server/Annotations/LastModified.php
+++ b/src/Server/Annotations/LastModified.php
@@ -5,7 +5,8 @@ declare(strict_types=1);
 namespace Laravel\Mcp\Server\Annotations;
 
 use Attribute;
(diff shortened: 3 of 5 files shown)


pushpak1300:
@taylorotwell  It could be, and it already works since the user can override the annotations method and return an array. But for defining [tool annotations](https://laravel.com/docs/10.x/mcp#tool-annotations) we already use attributes, so it would be better if the codebase supported this in a consistent way instead of having two completely different approaches.

PR #113, opened 2025-11-24
Add Support For Resource Templates

### LINKED ISSUE #69: Support resource templates
https://modelcontextprotocol.io/specification/2025-06-18/server/resources#resource-templates

pushpak1300:
Resolves #69

This PR introduces support for [ResourceTemplate](https://modelcontextprotocol.io/specification/2025-06-18/server/resources#resource-templates) using an interface based approach. The earlier implementation required extending a dedicated ResourceTemplate class, which added extra structure that was not really necessary. With this update, URI template support becomes an optional capability that any resource can opt into by implementing a small interface.

## Usage

```php
use Laravel\Mcp\Server\Contracts\SupportsURITemplate;
use Laravel\Mcp\Server\Resource;
use Laravel\Mcp\Support\UriTemplate;

class UserFileResource extends Resource implements SupportsURITemplate
{
    protected string $mimeType = 'text/plain';

    protected string $description = 'Access user files by ID';

    public function uriTemplate(): UriTemplate
    {
        return new UriTemplate('file://users/{userId}/files/{fileId}');
    }

    { 
        $userId = $request->get('userId');
        $fileId = $request->get('fileId');
        $uri = $request->uri();

        return Response::text("User $userId, File $fileId");
    }
}
```

### Why Simplified

The MCP spec references RFC 6570 for URI template behavior, but full RFC support is more than what MCP resources typically need. This intentionally implements a simpler subset. The approach mirrors what the official PHP MCP SDK docs [here](https://github.com/modelcontextprotocol/php-sdk/blob/59ab0888b1de09f8022e868c5379cf495da7ca41/src/Capability/Registry/ResourceTemplateReference.php#L217), and we followed the same interpretation. 

### Supported

* Simple placeholders like `{variable}`
* Multiple variables in a path such as `{var1}/{var2}`
* Variables match a single path segment
* Alphanumeric variable names

```php
new UriTemplate('file://users/{userId}');
new UriTemplate('file://users/{userId}/files/{fileId}');
new UriTemplate('https://api.example.com/{version}/{resource}/{id}');
```

### Not Supported

* `{+var}`
* `{#var}`
* `{.var}`
* `{/var}`
* `{?q,limit}`
* `{&page}`
* `{var*}`

> [!IMPORTANT]
>  Variables cannot span across multiple path segments. 
> ```php
> // Valid
> $template = new UriTemplate('file://files/{filename}');
> $template->match('file://files/document.txt'); // returns ['filename' => 'document.txt']
> // Not valid
> $template->match('file://files/folder/document.txt'); // returns null
> ```

### DIFF
--- a/src/Server/McpServiceProvider.php
+++ b/src/Server/McpServiceProvider.php
@@ -9,7 +9,6 @@ use Illuminate\Support\ServiceProvider;
 use Laravel\Mcp\Console\Commands\InspectorCommand;
 use Laravel\Mcp\Console\Commands\MakePromptCommand;
 use Laravel\Mcp\Console\Commands\MakeResourceCommand;
-use Laravel\Mcp\Console\Commands\MakeResourceTemplateCommand;
 use Laravel\Mcp\Console\Commands\MakeServerCommand;
 use Laravel\Mcp\Console\Commands\MakeToolCommand;
 use Laravel\Mcp\Console\Commands\StartCommand;
@@ -95,7 +94,6 @@ class McpServiceProvider extends ServiceProvider
             MakeToolCommand::class,
             MakePromptCommand::class,
             MakeResourceCommand::class,
-            MakeResourceTemplateCommand::class,
             InspectorCommand::class,
         ]);
     }


src/Server/Contracts/SupportsURITemplate.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laravel\Mcp\Server\Contracts;
+
+use Laravel\Mcp\Support\UriTemplate;
+
+interface SupportsURITemplate
```
joetannenbaum: I think it should be `SupportsUriTemplate`

src/Support/UriTemplate.php
```diff
@@ -0,0 +1,142 @@
...
+        if (! preg_match(self::URI_TEMPLATE_PATTERN, $template)) {
+            throw new InvalidArgumentException('Invalid URI template: must be a valid URI template with at least one placeholder.');
+        }
+
+        $this->variableNames = $this->extractVariableNames($template);
+    }
+
+    public static function make(string $template): self
+    {
+        return self::$instances[$template] ??= new UriTemplate($template);
```
joetannenbaum: Is it possible that multiple places are calling the same template? Is that why we're doing this?

src/Support/UriTemplate.php
```diff
@@ -0,0 +1,133 @@
...
+
+namespace Laravel\Mcp\Support;
+
+use Illuminate\Support\Str;
+use InvalidArgumentException;
+use Stringable;
+
+class UriTemplate implements Stringable
+{
+    private const MAX_TEMPLATE_LENGTH = 1000000;
```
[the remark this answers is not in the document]
pushpak1300: Yeah make sense. Thanks

PR #119, opened 2025-11-27
Add Logging Support

### LINKED ISSUE #118: Add Support For Logging
https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/logging

pushpak1300:
This PR adds support for the [MCP logging utility](https://modelcontextprotocol.io/specification/2025-11-25/server/utilities/logging), allowing servers to send structured log messages to clients.

Logging is **opt in**. MCP servers will not enable this by default. To turn it on, the user must explicitly add the capability in their MCP server class:

```php
protected function boot(): void
{
    $this->addCapability(self::CAPABILITY_LOGGING);
}
```

Once enabled, client can send `logging/setLevel` request. User can log things but filtering won't happen without this capability.

## How It Works

1. **Default level**: Info (Debug is filtered out unless enabled)
2. **Client control**: Clients call `logging/setLevel` to adjust verbosity
3. **Automatic filtering**: Server only emits logs at or above the current threshold
4. **Session isolation**: Each connection maintains its own log level
5. **Cache-backed state**: Session data is stored via Laravel cache, which works for both STDIO and HTTP transports (unlike Laravel sessions)

### Basic Logging

```php
use Laravel\Mcp\Enums\LogLevel;

class DatabaseTool extends Tool
{
    public function handle(Request $request): Generator
    {
        yield Response::log(LogLevel::INFO, 'Connecting to database');

        try {
            $result = DB::query(...);
            yield Response::log(LogLevel::DEBUG, ['query' => $sql, 'rows' => $result->count()]);
            yield Response::text("Found {$result->count()} records");
        } catch (\Exception $e) {
            yield Response::log(LogLevel::ERROR, ['error' => $e->getMessage()], 'database');
            yield Response::error('Query failed');
        }
    }
}
```

### Testing

```php
MyServer::tool(DatabaseTool::class)
    ->assertLogSent(LogLevel::INFO, 'Connecting')
    ->assertLogSent(LogLevel::ERROR)
    ->assertLogCount(2);
```

## Configuration

```php
// config/mcp.php
'session_ttl' => env('MCP_SESSION_TTL', 86400), // 24 hours
```


joetannenbaum (changes requested):
This looks good overall, just the one nit comment about the cases as opposed to hardcoding, I'll let you make the call there.

Is this the first time we're doing something like `$this->addCapability('logging');` as a public facing user API? Is that string essentially arbitrary or are there pre-defined use cases? Either way, for built-in capabilities, do we want to avoid the magic string and provide a helper method such as `enableLogging` or something similar?

src/Server/Support/SessionStore.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laravel\Mcp\Server\Support;
+
+use Illuminate\Contracts\Cache\Repository as Cache;
+
+class SessionStore
```
joetannenbaum, line 9: A whole SessionStore setup seems a bit overblown for this, we're not even using all of the methods. Are we planning on utilizing this in the future for something specific? If not, let's just have the logging manager class interact directly with the cache, right?

src/Server.php
```diff
@@ -99,6 +108,7 @@ abstract class Server
         'prompts/list' => ListPrompts::class,
         'prompts/get' => GetPrompt::class,
         'ping' => Ping::class,
+        'logging/setLevel' => SetLogLevel::class,
```
joetannenbaum, line 111: Shouldn't this only be here if the capability is enabled?

src/Server/Content/Log.php
```diff
@@ -0,0 +1,50 @@
...
+        protected LogLevel $level,
+        protected mixed $data,
+        protected ?string $logger = null,
+    ) {
+        parent::__construct('notifications/message', $this->buildParams());
+    }
+
+    public function level(): LogLevel
+    {
+        return $this->level;
```
joetannenbaum, line 21: If these are never set after the contructor, any reason to not make them just public readonly properties? Instead of getter methods?

joetannenbaum (changes requested):
Ok, well, just reviewed this not realizing it was a draft 🤦 

Regardless, if we decide to ever go with this please refer to my comments.

src/Server/Support/SessionStore.php
pushpak1300: Yeah that can be done.

src/Server.php
pushpak1300: this is reference for the methods. we are checking in handler if capability is enabled or not.

pushpak1300:
This is deprecated.

PR #120, opened 2025-11-28
Update JsonSchema usage

pushpak1300:
This PR updates the codebase to rely on the `JsonSchema` contract interface instead of the concrete implementation class. This improves the autocompletion.

> [!CAUTION]
>  **Warning: Breaking Change**
>
> Tool implementations that explicitly type hint `Illuminate\JsonSchema\JsonSchema` in their `schema()` or `outputSchema()` methods must update to use the contract interface `Illuminate\Contracts\JsonSchema\JsonSchema`.
>
>### Migration Guide
>
> ```php
> // Before
> use Illuminate\JsonSchema\JsonSchema;
>
> public function schema(JsonSchema $schema): array
> {
>    //
> }
> ```
>
> ```php
> // After
> use Illuminate\Contracts\JsonSchema\JsonSchema;
>
> public function schema(JsonSchema $schema): array
> {
>    //
> }
> ```
>
> This affects only custom tool classes that override the schema methods. The update is minimal, limited to changing the import to the contract interface.

### DIFF
--- a/tests/Feature/Console/StartCommandTest.php
+++ b/tests/Feature/Console/StartCommandTest.php
@@ -105,8 +105,7 @@ it('can stream a tool response over http', function (): void {
 
     $response->assertStatus(200);
 
-    $headers = $response->headers;
-    $headers->contains('Content-Type', 'text/event-stream');
+    expect(strtolower((string) $response->headers->get('Content-Type')))->toBe('text/event-stream; charset=utf-8');
 
     $content = $response->streamedContent();
     $messages = parseJsonRpcMessagesFromSseStream($content);


tests/Feature/Console/StartCommandTest.php
```diff
@@ -104,7 +104,8 @@
     );
 
     $response->assertStatus(200);
-    $response->assertHeader('Content-Type', 'text/event-stream; charset=UTF-8');
+
+    expect(strtolower((string) $response->headers->get('Content-Type')))->toBe('text/event-stream; charset=utf-8');
```
pushpak1300: This test started failing on the Laravel 11 and 12 because of `UTF-8`case so just added test to test this way.

PR #146, opened 2026-02-08
Update command description formatting to match all other Laravel Commands

matthewjohns0n:
Removed the period from the command description. This is consistent with all other Laravel commands, which do not end in a period.


pushpak1300:
Thanks !

PR #189, opened 2026-04-05
fix: validate OAuth registration client names

Gujiassh:
## Summary
- require OAuth client registration requests to provide either `client_name` or the legacy `name` field
- normalize the controller to prefer `client_name` while still falling back to `name` for older clients
- add focused registrar coverage for missing names, legacy fallback, and `client_name` precedence

## Testing
- `vendor/bin/pest tests/Unit/Server/RegistrarTest.php`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,27 @@
 # Release Notes
 
-## [Unreleased](https://github.com/laravel/mcp/compare/v0.6.3...main)
+## [Unreleased](https://github.com/laravel/mcp/compare/v0.6.5...main)
+
+## [v0.6.5](https://github.com/laravel/mcp/compare/v0.6.4...v0.6.5) - 2026-03-30
+
+### What's Changed
+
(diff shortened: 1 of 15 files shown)


Gujiassh:
Refreshed this branch onto the latest `main` and reverified the merged branch tip locally.

Local verification after the refresh:
- `composer test:lint` → passed
- `php vendor/bin/pest tests/Unit/Server/RegistrarTest.php` → passed
- `php vendor/bin/pest tests/Feature/Testing/Tools/AssertStructuredContentTest.php` → passed
- `php vendor/bin/pest --ci` → passed
- `php vendor/bin/phpstan --no-progress` → passed

Note: `composer test` still shells into coverage mode (`pest --coverage`), and this machine doesn't currently have a coverage driver installed, so I used the equivalent non-coverage Pest run plus phpstan for local verification. The OAuth client-name validation behavior is still intact on the refreshed branch.

Gujiassh:
Follow-up: the refreshed branch is now pushed on the PR head ref (`fix/oauth-client-name-validation-clean`). GitHub still appears to be showing the old head SHA immediately after the push, so the current `DIRTY` state looks like UI/cache lag rather than an unpushed branch refresh.

PR #192, opened 2026-04-11
Preserve existing OAuth discovery routes

Gujiassh:
When an application registers its own `/.well-known/oauth-authorization-server` or `/.well-known/oauth-protected-resource` routes before calling `oauthRoutes()`, the MCP package unconditionally overwrites them with its own handlers. This means apps that need custom OAuth discovery metadata (e.g. a separate authorization server) lose their routes silently.

This PR detects pre-existing exact GET routes at those well-known paths and skips registering the MCP handler when one already exists. Nested discovery (e.g. `/.well-known/oauth-authorization-server/mcp/weather`) continues to work on a separate `{path}` route that is always registered.

Also fixes the `Mcp` facade docblock types for PHPStan and updates `AddWwwAuthenticateHeader` middleware to reference the always-present `.nested` route name.

Supersedes #179.

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -10,6 +10,3 @@ parameters:
   ignoreErrors:
     - "#Unsafe usage of new static\\(\\)#"
     - "#Call to an undefined method Illuminate\\\\Contracts\\\\Foundation\\\\Application::routesAreCached#"
-    -
-      message: "#no value type specified in iterable type array#"
-      path: src/Facades/Mcp.php
--- a/src/Facades/Mcp.php
+++ b/src/Facades/Mcp.php
@@ -8,8 +8,8 @@ use Illuminate\Support\Facades\Facade;
 use Laravel\Mcp\Server\Registrar;
 
 /**
- * @method static \Illuminate\Routing\Route web(string $route, string<\Laravel\Mcp\Server> $serverClass)
- * @method static void local(string $handle, string<\Laravel\Mcp\Server> $serverClass)
+ * @method static \Illuminate\Routing\Route web(string $route, class-string<\Laravel\Mcp\Server> $serverClass)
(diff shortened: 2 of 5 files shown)


Gujiassh:
Maintenance refresh on current head branch.

Validation run:
- ./vendor/bin/pest tests/Unit/Server/RegistrarTest.php

Result:
- pass (41 passed).

pushpak1300:
Can you pleas explain the issue you're facing. It's hard to understand whats exactly bug and if there is way to reproduce it.

Gujiassh:
Repro / issue summary:

If an app already defines the exact well-known endpoints at the root, for example `/.well-known/oauth-protected-resource` or `/.well-known/oauth-authorization-server`, calling `Mcp::oauthRoutes()` currently registers the optional `{path?}` discovery routes in a way that takes over those exact endpoints too.

That breaks two cases:
- the app can no longer keep its own exact root metadata response
- nested discovery like `/.well-known/oauth-protected-resource/mcp/weather` still needs to work for resource-specific metadata

This PR changes the registrar so it only adds the exact root routes when they do not already exist, while still always registering the nested `{path}` variants for sub-resources. The added tests cover both behaviors.

I also pushed a small follow-up commit (`d3b62d6`) that fixes an unrelated `phpstan` failure on the `Mcp` facade docblock so CI can run green again.

Local verification on the current branch tip:
- `./vendor/bin/pest tests/Unit/Server/RegistrarTest.php`
- `./vendor/bin/phpstan analyse --no-progress`

pushpak1300:
## Auto code review

No issues found. Checked for bugs and CLAUDE.md compliance.

pushpak1300:
I have updated the description and also remove the nested level issue for different path. it will use root level issue for nested path now.

pushpak1300 (approved):

PR #196, opened 2026-04-14
Add MCP UI App support

pushpak1300:
MCP currently has no way to show interactive UI to users — every tool response is plain text. This adds MCP App support: a tool linked to an `AppResource` can render a self-contained HTML app inside a sandboxed iframe in the host.

### Usage

```bash
php artisan make:mcp-app-resource DashboardApp
```

This generates two files. The PHP class registers the resource — the default `handle()` auto-infers the Blade view from the class name, so no code is needed unless you're passing server-side data:

```php
class DashboardApp extends AppResource
{
    {
        return Response::view('mcp.dashboard-app', [
            'title' => $this->title(),
        ]);
    }
}
```

The Blade view is the entire app — HTML, JS, and CSS in one file:

```blade
<x-mcp::app title="Dashboard">
    <x-slot:head>
        <script type="module">
        createMcpApp(async (app) => {
            const result = await app.callTool('get-data');
            document.getElementById('output').textContent = result.content[0]?.text;
        });
        </script>
    </x-slot:head>
    <div id="output"></div>
</x-mcp::app>
```

Link a tool to display the app when called:

```php
#[RendersApp(resource: DashboardApp::class)]
class ShowDashboard extends Tool
{
    {
        return Response::text('Dashboard loaded.');
    }
}
```

App-only tools (invisible to the model, callable only from the UI):

```php
#[RendersApp(resource: DashboardApp::class, visibility: ['app'])]
class GetDashboardData extends Tool { ... }
```

### Approach

- `AppResource` extends `Resource` with `ui://` URI scheme and `text/html;profile=mcp-app` MIME type; serves self-contained HTML via `handle()` which renders a Blade view
- `#[RendersApp]` attribute on a tool injects `_meta.ui.resourceUri` into its tool listing so the host knows which resource to fetch
- The `x-mcp::app` Blade component inlines the pre-bundled JS SDK so apps require no npm or Vite
- `AppMeta` / `Csp` / `Permissions` provide structured configuration for iframe sandbox permissions and CSP
- Server advertises `io.modelcontextprotocol/ui` capability automatically when any `AppResource` is registered


wojo1206:
This is large addition to project. AI generated?

pushpak1300:
Yes @wojo1206

wojo1206:
@pushpak1300 Documentation is 1K and the `resources/js/mcp-sdk.js` is .5K. Is `mcp-sdk.js` standardized or invented by your AI?

resources/js/mcp-sdk.js
```diff
@@ -0,0 +1,488 @@
...
+                params.logger = logger;
+            }
+
+            notify('notifications/message', params);
+
+            return Promise.resolve();
+        }
+
+        await setup({
+            hostContext: function () {
```
joetannenbaum: Why all of the aliases here? We don't need double functions everywhere unless there's a good reason.

resources/js/mcp-sdk.js
```diff
@@ -0,0 +1,488 @@
...
+                    data: data,
+                };
+
+            if (!isObject(levelOrParams) && logger !== undefined) {
+                params.logger = logger;
+            }
+
+            notify('notifications/message', params);
+
+            return Promise.resolve();
```
joetannenbaum, line 488: Why `Promise.resolve()` here?

resources/js/mcp-sdk.js
```diff
@@ -0,0 +1,488 @@
...
+        const handler = requestHandlers[message.method];
+
+        if (handler) {
+            handler();
+        } else {
+            respondError(message.id, -32601, 'Method not found: ' + message.method);
+        }
+    }
+
+    window.addEventListener('message', function (event) {
```
joetannenbaum: We should clean this up, right?

resources/js/mcp-sdk.js
```diff
@@ -0,0 +1,488 @@
...
+    function notifySizeChanged() {
+        notify('ui/notifications/size-changed', currentSize());
+    }
+
+    function autoResize() {
+        if (typeof ResizeObserver === 'undefined' || !document.body) {
+            return;
+        }
+
+        const observer = new ResizeObserver(notifySizeChanged);
```
joetannenbaum: We should clean up this observer, right?

src/Server/Ui/Permissions.php
```diff
@@ -0,0 +1,87 @@
...
+    {
+        return array_map(
+            fn (): stdClass => new stdClass,
+            array_filter([
+                'camera' => $this->camera,
+                'microphone' => $this->microphone,
+                'geolocation' => $this->geolocation,
+                'clipboardWrite' => $this->clipboardWrite,
+            ]),
+        );
```
joetannenbaum: Seems like we're doing this in order to get "{}" after json encoding... why do we want an empty object?

resources/js/mcp-sdk.js
```diff
@@ -0,0 +1,534 @@
+(function () {
+    const jsonRpcVersion = "2.0";
+    const protocolVersion = "2026-01-26";
```
joetannenbaum, line 3: Just want to confirm we don't need to align this protocol with anything on the PHP side

joetannenbaum (commented):
Overall good, made some adjustments here and there, have a couple of questions. Let me know what you think.

If I changed anything that's fundamentally against your specific design or intention let me know.

resources/js/mcp-sdk.js
pushpak1300: I kept it duplicate for the consistency with the official sdk and kept aliases for good DX. if doesn't make sense i can remove the duplicate one.

resources/js/mcp-sdk.js
pushpak1300: For the consistency even though `sendLog` is kind of fire and forget thing all the function are async in the sdk so wanted to keep the consistency for the end user while writing code and also the official sdk follow he same approach.

`await app.sendLog('info', 'msg')`

resources/js/mcp-sdk.js
pushpak1300: Yes make sense. will clean it up.

resources/js/mcp-sdk.js
pushpak1300: https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx It's different and separate version.

src/Server/Ui/Permissions.php
pushpak1300: yeah it's little weird to me as well about this but this is how thy have mentioned in resource format protocol. https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx#ui-resource-format

resources/js/mcp-sdk.js
pushpak1300: Yes this needs to be clean.

pushpak1300:
> @pushpak1300 Documentation is 1K and the resources/js/mcp-sdk.js is .5K. Is mcp-sdk.js standardized or invented by your AI?

@wojo1206 which documentation ?

resources/js/mcp-sdk.js
joetannenbaum: Yeah I don't think we need multiple aliases for everything, let's just stick with the official SDK. Easy to layer in later if we want, but harder to remove once it's in there.

src/Server.php
```diff
@@ -282,11 +286,13 @@ protected function runMethodHandle(JsonRpcRequest $request, ServerContext $conte
         );
 
         $container->instance('mcp.request', $request->toRequest());
+        $container->instance('mcp.current_server', $this);
```
joetannenbaum: Is this ever used?

joetannenbaum (commented):
Just the one question, looks good otherwise.

vinitkadam03:
This is awesome, exactly what we needed. Really excited about it, hope the team can ship it soon. Definitely looking forward to using this!

stuartbrameld:
Agreed. I was really concerned Laravel was behind the times a year ago, and my business depended on it. All that has changed with the push into AI. This kind of thing is fantastic.

PR #197, opened 2026-04-14
Fix route:cache memory exhaustion caused by circular reference

### LINKED ISSUE #198: route:cache fails with infinite recursion
### Laravel Package Version

0.6.6

### Laravel Version

12.56.0

### PHP Version

8.4.20

### Database Driver & Version

_No response_

### Description

MCP v0.6.6 breaks `php artisan route:cache` with an infinite recursion error. The cause is PR #192 ("Preserve existing OAuth discovery routes"), which refactored the OAuth discovery route closures in `Registrar::oauthRoutes()` to capture `$this`.

In v0.6.5, the closures used inline arrays and global helpers (url(), config()), so no object references were captured:

```php
Router::get('/.well-known/oauth-protected-resource/{path?}', fn (?string $path = '') => response()->json([
    'resource' => url('/'.$path),
    'authorization_servers' => [config('mcp.authorization_server') ?? url('/')],
    'scopes_supported' => ['mcp:use'],
]));
```

In v0.6.6, the closures now capture `$this` (the Registrar instance) via method calls:

```php
Router::get('/.well-known/oauth-protected-resource', fn () => response()->json($this->protectedResourceMetadata('')));
```

The Registrar holds Route objects in `$this->httpServers`, which reference the application container. The container holds singletons with closure properties, creating circular references that cause `laravel/serializable-closure` to hit infinite recursion during route serialization.

Rolling back to v0.6.5 resolves the issue.

### Steps To Reproduce

1. Install `laravel/mcp:v0.6.6` with `Mcp::oauthRoutes()` in `routes/ai.php`
2. Register at least one MCP server with `Mcp::web()`
3. Run `php artisan route:cache`
4. Observe infinite recursion error in `laravel/serializable-closure` at `vendor/laravel/serializable-closure/src/Serializers/Native.php:567`
Image

5. Downgrade to laravel/mcp:v0.6.5 -- route:cache succeeds

radumetes:
## Fix `route:cache` memory exhaustion

Running `php artisan route:cache` (or `optimize`) with MCP routes registered causes a fatal memory exhaustion error — even with 2GB of memory. This is due to a circular reference during route serialization.

### The Problem

The `Registrar` stores `Route` objects on `$this->httpServers`. When `route:cache` serializes these routes, it walks into closures that reference the `Registrar`, which holds the `Route` objects, which contain the closures — infinite recursion.

The `oauthRoutes()` closures also capture `$this` via instance method calls (`$this->protectedResourceMetadata()`), creating the same cycle.

### The Fix

- Store URI strings instead of `Route` objects on the `Registrar`
- Resolve `Route` objects from the router on demand in `getWebServer()`
- Make metadata methods `static` so closures don't capture `$this`

### No Breaking Changes

The public API is unchanged — `getWebServer()` still returns `?Route` and `servers()` returns the same shape. All 641 existing tests pass without modification.

### DIFF
--- a/src/Console/Commands/InspectorCommand.php
+++ b/src/Console/Commands/InspectorCommand.php
@@ -6,7 +6,6 @@ namespace Laravel\Mcp\Console\Commands;
 
 use Exception;
 use Illuminate\Console\Command;
-use Illuminate\Routing\Route;
 use Illuminate\Support\Arr;
 use Laravel\Mcp\Server\Registrar;
 use Symfony\Component\Console\Attribute\AsCommand;
@@ -48,7 +47,7 @@ class InspectorCommand extends Command
             $server = array_shift($servers);
             [$localServer, $route] = match (true) {
                 is_callable($server) => [$server, null],
-                $server::class === Route::class => [null, $server],
+                is_string($server) => [null, $registrar->getWebServer($server)],
                 default => [null, null],
             };
         }
--- a/src/Console/Commands/InspectorCommand.php
+++ b/src/Console/Commands/InspectorCommand.php
@@ -6,6 +6,7 @@ namespace Laravel\Mcp\Console\Commands;
 
 use Exception;
 use Illuminate\Console\Command;
+use Illuminate\Routing\Route;
 use Illuminate\Support\Arr;
 use Laravel\Mcp\Server\Registrar;
 use Symfony\Component\Console\Attribute\AsCommand;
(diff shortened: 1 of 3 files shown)


wojo1206:
I observed the problem and tracked this down to `Mcp::oauthRoutes();`. Without that call, my `php artisan route:cache` completes fine. Your solution doesn't seem to work.

radumetes:
@wojo1206 the fix works for all combinations — oauthRoutes() only, web() only, and both together. Tested locally with php -d memory_limit=512M artisan route:cache. Could you confirm you're using the full changeset from this PR? The fix requires all changes in Registrar.php applied together.

wojo1206:
@radumetes Yep, you are right. I just didn't use right branch. I just tried `php artisan route:cache` when `Mcp::oauthRoutes()` and your solutions works. It is funny that I observed that issue and just 15min later your merge request appeared ;)

pushpak1300:
Thanks.

radumetes:
Happy to help!

PR #205, opened 2026-04-30
Add InvokingTool and ToolInvoked events

radumetes:
Adds a pre/post invocation event pair around tool calls, mirroring the pattern already used by `laravel/ai` (`InvokingTool` / `ToolInvoked`). Listeners can hook tool invocations for audit logs, metrics, rate limiting, and similar cross-cutting concerns without subclassing every `Tool`.

## Behaviour

- `InvokingTool` fires before the tool's `handle()` runs.
- `ToolInvoked` fires after, on successful responses and on `AuthenticationException` / `AuthorizationException` / `ValidationException` that `CallTool` already converts to error responses.
- Uncaught exceptions propagate without firing `ToolInvoked`, matching `laravel/ai`'s convention.
- The shared `invocationId` correlates the two events.

## Streamed tools

For tools that return a `Generator`, a passthrough generator collects yielded responses as they're consumed by the framework and dispatches `ToolInvoked` from a `finally` block once iteration completes. Listeners receive the full collected response set without consuming the stream the framework iterates.

## Usage

```php
use Laravel\Mcp\Events\ToolInvoked;

Event::listen(function (ToolInvoked $event): void {
    activity('user')
        ->causedBy($event->request->user())
        ->event($event->tool->name())
        ->withProperties([
            'tool'       => $event->tool->name(),
            'arguments'  => $event->request->all(),
            'session_id' => $event->request->sessionId(),
        ])
        ->log('MCP '.$event->tool->name());
});
```

Implement `ShouldQueue` on the listener to push the work off the request thread.

## Compatibility

Purely additive, no existing behaviour changes.

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -10,6 +10,3 @@ parameters:
   ignoreErrors:
     - "#Unsafe usage of new static\\(\\)#"
     - "#Call to an undefined method Illuminate\\\\Contracts\\\\Foundation\\\\Application::routesAreCached#"
-    -
-      message: "#no value type specified in iterable type array#"
-      path: src/Facades/Mcp.php
--- a/src/Events/InvokingTool.php
+++ b/src/Events/InvokingTool.php
@@ -1,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laravel\Mcp\Events;
+
+use Laravel\Mcp\Request;
+use Laravel\Mcp\Server\Tool;
+
+class InvokingTool
+{
+    public function __construct(
+        public string $invocationId,
+        public Tool $tool,
+        public Request $request,
+    ) {
+        //
+    }
+}
--- a/src/Events/ToolInvoked.php
+++ b/src/Events/ToolInvoked.php
@@ -1,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laravel\Mcp\Events;
+
+use Laravel\Mcp\Request;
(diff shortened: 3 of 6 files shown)


radumetes:
CI lint fails on src/Facades/Mcp.php with fully_qualified_strict_types, this also fails on main with no changes, unrelated to this PR.

pushpak1300:
@radumetes can you fix the phpstan errors. ?

radumetes:
@pushpak1300 Added array<...> value types to @method servers() and @method ensureMcpScope() in Mcp.php, which resolves the underlying missingType.iterableValue errors. Also removed the related ignore rule in phpstan.neon.dist (added by #207) since it now triggers ignore.unmatched once the errors are fixed.

taylorotwell:
The submission doesn't mention a real world need driven by real application development. Only hypothetical consistency with laravel/ai.

radumetes:
@taylorotwell I run a multi-tenant SaaS that exposes an MCP server, every tool call has to be audit logged for compliance (OAuth client, org, session, args, duration, status, result, usage). The only way I found to keep it consistent today is an abstract LoggedTool base class with a final handle() so nothing bypasses it, an execute() method tools implement instead, a try/catch/finally so the audit still writes when a tool throws, and a resultMetaForLogging() hook each tool has to override for its own metrics. Forget that override and you silently lose half the audit.

These events make the whole class go away. One listener handles it, new tools extend Tool like normal, logging is automatic.

PR #207, opened 2026-05-05
Return 201 from OAuth client registration per RFC 7591

danniehansen:
This PR adjusts the response code of the OAuth client registration to match the spec https://www.rfc-editor.org/rfc/rfc7591.html#section-3.2

Today it's responding with 200 while spec specifies 201.

I ran into this issue when attempting to use our MCP server in Antigravity which seems to be more strict when it comes to spec alignment. Where-as ChatGPT and others seem to allow 200 just the same.

Adjusted unit tests to match the new expected status code.

Not sure in terms of compatibility. Some users may rely on status code 200, even though the spec dictates 201.

### DIFF
--- a/.github/workflows/facade.yml
+++ b/.github/workflows/facade.yml
@@ -39,6 +39,9 @@ jobs:
       - name: Update facade docblocks
         run: php -f vendor/bin/facade.php -- Laravel\\Mcp\\Facades\\Mcp
 
+      - name: Format with Pint
+        run: vendor/bin/pint src/Facades/Mcp.php
+
       - name: Commit facade docblocks
         uses: stefanzweifel/git-auto-commit-action@v5
         with:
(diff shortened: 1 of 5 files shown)


.github/workflows/facade.yml
```diff
@@ -39,6 +39,9 @@ jobs:
       - name: Update facade docblocks
         run: php -f vendor/bin/facade.php -- Laravel\\Mcp\\Facades\\Mcp
 
+      - name: Format with Pint
+        run: vendor/bin/pint src/Facades/Mcp.php
+
```
pushpak1300, line 44: Added this to avoid formatting conflicts with pint and Facade documentor.

pushpak1300 (approved):

pushpak1300:
Thanks.

PR #216, opened 2026-05-14
Add MCP client foundation with stdio transport and ping

pushpak1300:
There's no MCP client in `laravel/mcp` yet. The package only ships the server side. This PR lands the smallest possible foundation: a builder API that can spawn a stdio MCP server, do the `initialize` handshake, and `ping` it.

```php
use Laravel\Mcp\Client;

$client = Client::local('/Users/pushpakchhajed/.bun/bin/bunx', ['-y', '@modelcontextprotocol/server-everything']);

try {
   $response = $client->connect();
   dump($response);
} finally {
    $client->disconnect();
}
```

Resulted into 

Screenshot 2026-05-20 at 2 07 09 AM

`ping()` lazily connects on first call, so the explicit `connect()` is optional.

Intentionally tiny so it's easy to review and merge without breaking anything. Tools/resources/prompts, HTTP transport, OAuth, the `Mcp::client()` facade, etc. all follow in later PRs per the shipping plan.

Stacked on #219 (the `ProtocolVersion` enum). `Initialize` sends `ProtocolVersion::LATEST->value`. Once #219 is in I'll rebase this onto main.

> [!NOTE]
> The public API is subject to change until release.


src/Schema/Implementation.php
```diff
@@ -36,4 +37,31 @@ public function toArray(): array
...
+        return new self(
+            name: Arr::get($data, 'name'),
+            version: Arr::get($data, 'version'),
+            title: Arr::get($data, 'title'),
+            description: Arr::get($data, 'description'),
+            icons: Arr::map(Arr::get($data, 'icons', []), fn (array $icon): Icon => Icon::from(
+                src: Arr::get($icon, 'src'),
+                mimeType: Arr::get($icon, 'mimeType'),
+                sizes: Arr::get($icon, 'sizes'),
+                theme: IconTheme::from(Arr::get($icon, 'theme'))
```
benbjurstrom: This fails when an icon omits the theme or sizes fields

src/Client.php
```diff
@@ -0,0 +1,203 @@
...
+    public function withTimeout(float $seconds): static
+    {
+        $this->transport->setTimeoutSeconds($seconds);
+
+        return $this;
+    }
+
+    public function connect(): static
+    {
+        if ($this->connected) {
```
WendellAdriel, line 69: I think we should reset the client state when a request fails after the handshake. Right now, this is trusting the `$connected` flag, but if the stdio transport times out, exits, or disconnects during a later call, the transport can be closed while the client still reports itself as connected. The next call will skip the handshake and fail with `Transport is not connected.`

Can we clear the connected state in that failure path and add a small regression test for: connect successfully, make the next request fail, then assert `connected()` is false?

composer.json
```diff
@@ -28,7 +28,8 @@
         "illuminate/routing": "^11.45.3|^12.41.1|^13.0",
         "illuminate/support": "^11.45.3|^12.41.1|^13.0",
         "illuminate/validation": "^11.45.3|^12.41.1|^13.0",
-        "illuminate/json-schema": "^12.41.1|^13.0"
+        "illuminate/json-schema": "^12.41.1|^13.0",
+        "illuminate/process": "^11.45.3|^12.41.1|^13.0"
```
WendellAdriel: This adds `illuminate/process`, but as far as I could see, you're using `Symfony\Component\Process` directly. I think we should either require `symfony/process` directly or switch the implementation to Illuminate's process API.

WendellAdriel (changes requested):

WendellAdriel (approved):
LGTM

josephkerkhof:
thank you @pushpak1300 for your work on this 🙇🏻

PR #219, opened 2026-05-18
Add Icon and Implementation (MCP spec 2025-11-25)

pushpak1300:
MCP clients currently cannot show visual identifiers for servers, tools, resources, or prompts because Laravel MCP does not expose icon metadata yet. MCP spec `2025-11-25` adds an `icons` field to `Implementation`, `Tool`, `Resource`, and `Prompt`.

This PR adds icon support across Laravel MCP.

It introduces:

- `Icon` value object
- `IconTheme` enum
- `#[Icon]` attribute
- `icons()` method hook

Icons can now be declared on `Server`, `Tool`, `Resource`, and `Prompt`.

```php
use Laravel\Mcp\Enums\IconTheme;
use Laravel\Mcp\Server\Attributes\Icon;

#[Icon('mcp/server.png', mimeType: 'image/png', sizes: ['48x48'])]
#[Icon('mcp/server-dark.svg', theme: IconTheme::Dark)]
class WeatherServer extends Server {}
````

You can also define icons programmatically:

```php
use Laravel\Mcp\Schema\Icon;

class WeatherTool extends Tool
{
    public function icons(): array
    {
        return [
            Icon::from('mcp/tool.png', mimeType: 'image/png'),
        ];
    }
}
```

Attributes and `icons()` can be used together. Both sources are combined automatically.

Icon paths are resolved as follows:

* URLs with a URI scheme, such as `https:` or `data:`, are used as-is
* Relative paths are resolved using Laravel’s `asset()` helper
* For full control, icons can be created directly with `new Icon(...)`

This PR also adds an `Implementation` value object for the MCP spec’s `Implementation` type. This will be shared by both the server and the upcoming client implementation.

`Initialize` now emits the full `Implementation` payload, including:

* `name`
* `version`
* `title`
* `description`
* `icons`
* `websiteUrl`

This is emitted regardless of the negotiated protocol version. Older clients safely ignore unknown fields.

> [!WARNING]
> **Breaking change in `ServerContext`.**
>
> `$context->serverName` and `$context->serverVersion` have been removed.
>
> `ServerContext` now exposes a single `Implementation` instance:
>
> ```php
> $context->implementation->name
> $context->implementation->version
> ```
>
> Constructor usage also changes:
>
> ```php
> // Before
> new ServerContext(
>     serverName: 'X',
>     serverVersion: 'Y',
> )
>
> // After
> new ServerContext(
>     implementation: new Implementation('X', 'Y'),
> )
> ```
>
> This should only affect custom `Method` handlers that read these properties directly. Most userland code should not need changes.


src/Schema/Icon.php
```diff
@@ -0,0 +1,88 @@
...
+    ) {
+        $scheme = strtolower((string) parse_url($src, PHP_URL_SCHEME));
+
+        if (! in_array($scheme, ['https', 'data'], true)) {
+            throw new InvalidArgumentException(
+                "Icon src must use https: or data: scheme, got: '{$scheme}'.",
+            );
+        }
+
+        if ($theme !== null && ! in_array($theme, ['light', 'dark'], true)) {
```
WendellAdriel: Wouldn't it be better to have `$theme` as an Enum instead?

WendellAdriel (changes requested):
Just added a small comment, but check if it makes sense.
Other than that, I think that an `Icon` attribute would be nice as well so we could do something like

```php
#[Icon('mcp/server.png', mimeType: 'image/png', sizes: ['48x48'])]
class WeatherServer extends Server
{
    // ...
}
```

PR #226, opened 2026-05-22
Add MCP client tool listing and calling

pushpak1300:
Currently the MCP client (from #216) can only connect, initialize, and ping. There's no way to discover or invoke the tools a server exposes. This PR adds that on top of the stdio foundation.

### Usage

```php
use Laravel\Mcp\Client;

$client = Client::local('npx', ['-y', '@modelcontextprotocol/server-everything']);

// List the server's tools (auto-paginated), keyed by name
$tools = $client->tools();
$tools->keys()->all();   // ['add', 'printEnv', ...]
$tools->has('add');      // true

// Cap how many you pull back
$tools = $client->tools(15);

// Call a tool straight off the listing
$result = $tools['add']->call(['a' => 1, 'b' => 2]);
$result->text();         // "3"
$result->isError;        // false

// Or call by name when you already know it
$result = $client->callTool('add', ['a' => 1, 'b' => 2]);
```

`tools()` returns a `Collection<string, Tool>` keyed by tool name, so lookups read naturally. `callTool()` and `$tool->call()` return a `ToolResult` value object (`->text()`, `->isError`, `->content`, plus `->structuredContent` and `->meta` when the server sends them). Application-level tool errors come back as a `ToolResult` with `isError === true`. Protocol errors throw `JsonRpcException`.


src/Client/Schema/ToolResult.php
```diff
@@ -0,0 +1,63 @@
...
+        public ?array $structuredContent = null,
+        public ?array $meta = null,
+    ) {
+        //
+    }
+
+    /**
+     * @param  array<string, mixed>  $result
+     */
+    public static function from(array $result): self
```
WendellAdriel, line 30: IMO, this should validate the `tools/call` result shape before creating the `ToolResult`. Right now malformed server payloads like a non-array `content` or non-boolean `isError` get converted into an empty successful result, which could make debugging harder. I think we can throw a `ClientException` for invalid payloads and add regression tests for those cases.

src/Client/Primitives/Tool.php
```diff
@@ -0,0 +1,63 @@
...
+        public readonly array $annotations,
+        public readonly ?array $meta,
+    ) {
+        //
+    }
+
+    /**
+     * @param  array<string, mixed>  $payload
+     */
+    public static function from(Client $client, array $payload): self
```
WendellAdriel, line 36: This value object is receiving untrusted server data, so I think we should validate before calling the `new self`. I think you can add more checks like you do with name and throw that same exception if anything is wrong/not expected.

src/Client/Schema/InitializeResult.php
```diff
@@ -26,26 +28,28 @@ public function __construct(
...
         }
 
-        $instructions = $payload['instructions'] ?? null;
-
         return new self(
             protocolVersion: $protocolVersion,
             capabilities: $capabilities,
             serverInfo: Implementation::from($serverInfo),
-            instructions: is_string($instructions) ? $instructions : null,
+            instructions: $instructions,
```
WendellAdriel: IDK if it's expected, but `instructions` used to be guarded as a string.

tests/Unit/Client/ToolsTest.php
```diff
@@ -0,0 +1,298 @@
+<?php
```
WendellAdriel, line 1: Maybe could be worth adding a few malformed-payload tests around `ToolResult::from()` and `Tool::from()` so these stay more strict when receiving bad data.

WendellAdriel (changes requested):

src/Client/Schema/InitializeResult.php
pushpak1300: As per specifications its always expected to be string form the server. so at client side not sure f we should do the validation because as per the MCP specs we need to be acceptable for all the things.

WendellAdriel (approved):
LGTM! :fire:

PR #227, opened 2026-05-23
Add MCP client Streamable HTTP transport

pushpak1300:
Currently the MCP client can only talk to servers over stdio (`Client::local()`), so there's no way to connect to a remote or hosted MCP server over HTTP.

This adds a Streamable HTTP transport (MCP 2025-11-25):

```php
// before: stdio only
$client = Client::local('npx', ['-y', '@modelcontextprotocol/server-everything']);

// after: connect over HTTP
$client = Client::web('https://mcp.example.com/mcp');
$client->tools();

// with a static bearer token
Client::web('https://mcp.example.com/mcp')->withToken(env('REMOTE_TOKEN'));
```

It POSTs each JSON-RPC frame via the `Http` facade (so it's `Http::fake()` testable), handles both `application/json` and `text/event-stream` responses, captures and echoes `MCP-Session-Id`, sends `MCP-Protocol-Version`, re-initializes on a 404, and sends a best-effort DELETE on disconnect. `withToken()` lives on a new `WebClient` subclass since stdio clients have no auth concept. No OAuth yet, that's a later stage.

Known limitation: SSE responses are read in full before parsing rather than incrementally, so a server that sends a blocking server->client request mid-stream (say a ping it waits on) would hang until the timeout. Narrow for this client since it advertises no sampling/roots/elicitation capabilities, but I'd like to follow up with incremental streaming. Open to feedback on whether that should block this PR.

Stacked on #226.


pushpak1300:
Ready for review. Putting in draft until #226 is merged.

src/Client/Transport/HttpTransport.php
```diff
@@ -0,0 +1,215 @@
...
+        $headers = [
+            'Accept' => 'application/json, text/event-stream',
+        ];
+
+        if ($this->sessionId !== null) {
+            $headers['MCP-Session-Id'] = $this->sessionId;
+        }
+
+        if ($this->initialized) {
+            $headers['MCP-Protocol-Version'] = ProtocolVersion::LATEST->value;
```
WendellAdriel, line 131: This should use the protocol version negotiated during `initialize`, not always `ProtocolVersion::LATEST`. A server can return an older supported version, and then these later HTTP requests may be rejected. Store the negotiated version and use it for all subsequent requests.

src/Client/Transport/HttpTransport.php
```diff
@@ -0,0 +1,215 @@
...
+            throw new SessionExpiredException("Session expired. The server responded with HTTP 404 for endpoint [{$this->url}].");
+        }
+
+        if ($status < 200 || $status >= 300) {
+            $this->failWith("Unexpected HTTP status [{$status}] from endpoint [{$this->url}].");
+        }
+
+        $this->initialized = true;
+
+        $body = $response->body();
```
WendellAdriel: Reading the full SSE body before parsing means the client cannot answer server-initiated requests until the stream closes. That can deadlock if the server sends a ping and waits for the response before sending the final result. I think we can parse SSE frames incrementally or fail for interactive SSE streams.

src/Client/Transport/HttpTransport.php
```diff
@@ -0,0 +1,215 @@
...
+                ->post($this->url);
+        } catch (ConnectionException $connectionException) {
+            $this->failWith("HTTP request to [{$this->url}] failed: {$connectionException->getMessage()}");
+        }
+
+        $this->captureSessionId($response);
+
+        $status = $response->status();
+
+        if ($status === 404) {
```
WendellAdriel: This treats every 404 as session expiry, but the spec only requires reinitialization when the failed request included `Mcp-Session-Id`. A normal missing endpoint would be misdiagnosed and retried. Let's only throw `SessionExpiredException` when the request actually had a session id.

WendellAdriel: Also, here and on other lines, I'd use a constant (from Response) for HTTP status codes instead of "magic numbers" for the codes.

WendellAdriel (changes requested):

src/Client/Transport/HttpTransport.php
pushpak1300: So for the client functionality we will not be supporting the older version and thats why this is intentional maybe adding exception protocol not supported is right option.

WendellAdriel (approved):
LGTM :rocket:

TitasGailius (approved):
LGTM, great job! 🔥

PR #228, opened 2026-05-26
Add named MCP clients with tool list caching

pushpak1300:
Currently there's no way to register a named MCP client and reuse it across a request, and every `tools()` call goes back to the server. For integrations you talk to often (Notion, GitHub, etc.) that's a real cost.

You register the client once and resolve it by name:

```php
// service provider
Mcp::registerClientFor('notion', fn () =>
    Client::web('https://mcp.notion.com')->oauth(env('NOTION_CLIENT_ID'))
);

// anywhere — same instance per request, list cached across requests
Mcp::client('notion')->tools();   // fetches + caches (default TTL 1h)
Mcp::client('notion')->tools();   // cache hit
Mcp::client('notion')->clearCache();

// tune or disable per registration
Mcp::registerClientFor('dev', fn () => Client::local('php', ['mcp.php']), cacheTtl: 0);
```

Inline clients (`Client::web/local()`) keep their existing behaviour, no caching, no name. Caching lives on the client class as an internal policy (`asRegisteredClient($name, $ttl)`).

Stacked on top of #227.


src/Client/Methods/Tools/ListTools.php
```diff
@@ -42,17 +44,59 @@ public function params(): array
...
+        }
+
+        return $this->fetch($protocol);
+    }
+
+    /**
+     * @param  array<int, array<string, mixed>>  $payloads
+     * @return Collection<string, Tool>
+     */
+    protected function hydrate(array $payloads): Collection
```
WendellAdriel, line 74: Cached payloads can fail with a `TypeError` when the cached array contains a non-array entry. In that case we never reach `Tool::from()`, so the cache is not flushed/refetched.

Maybe we can validate the cached payload before hydrating, or normalize this path into a `ClientException` so the existing cache recovery works.

If you think this is a good idea, a new regression test with a cached non-array tool entry would be nice too.

src/Client/Methods/Tools/ListTools.php
```diff
@@ -42,17 +44,59 @@ public function params(): array
      * @return Collection<string, Tool>
      */
     public function handle(Protocol $protocol): Collection
+    {
+        try {
+            return $this->hydrate($this->loadPayloads($protocol));
+        } catch (ClientException $clientException) {
```
WendellAdriel: This retry catches every `ClientException`, including failures from a `tools/list` request. That means a bad server response or transport/client error gets retried and can send duplicate requests.

Maybe limit the flush/refetch behavior to failed hydration of cached payloads only, and let live fetch failures fail once?

src/Server/McpServiceProvider.php
```diff
@@ -21,6 +22,8 @@ public function register(): void
     {
         $this->app->singleton(Registrar::class, fn (): Registrar => new Registrar);
 
+        $this->app->singleton(ClientManager::class, fn (): ClientManager => new ClientManager);
```
WendellAdriel: This being a singleton can leak state in Octane or other long-lived workers, especially if the factory captures a user token.

Make the resolved clients truly request-scoped, or add explicit Octane lifecycle flushing similar to other Laravel packages that keep request-sensitive state.

src/Client/ClientManager.php
```diff
@@ -0,0 +1,67 @@
...
+    protected array $clients = [];
+
+    /**
+     * @param  Closure(): Client  $factory
+     * @param  ?Closure(): (string|int|Authenticatable|null)  $scope
+     */
+    public function registerClient(
+        string $name,
+        Closure $factory,
+        ?int $cache = null,
```
WendellAdriel: In the PR description we have `cache: false`, but the method accepts `?int $cache`. In strict user code that will be a `TypeError`.

IMO, we should either support `bool|int|null` explicitly and normalize it, or update name to `cacheTTL` and make `0` being "no cache".

src/Client.php
```diff
@@ -98,6 +124,31 @@ public function callTool(string $name, array $arguments = []): ToolResult
         return (new CallTool($name, $arguments))->handle($this->protocol);
     }
 
+    public function clearCache(): void
```
WendellAdriel: I know we discussed in the meeting, but while reading the PR, I thought about something.
Right now only tools are cached, but if we add more cache later, this `clearCache` should clear all the cache.
If we plan to add more cache data and want granular control, we may add another method called `clearTools` or something like this to clear just the tools cache.

src/Client.php
```diff
@@ -55,6 +69,18 @@ public function withTimeout(float $seconds): static
         return $this;
     }
 
+    /**
+     * @param  ?Closure(): (string|int|Authenticatable|null)  $scope
+     */
+    public function asRegisteredClient(string $name, int $cache, ?Closure $scope = null): static
```
WendellAdriel: Do we need this as public?

Because it exposes registration internals on the public `Client` API, where users can call it directly and mutate hidden cache/registry state, which is a bit weird and maybe we shouldn't allow users to touch.

If it needs to be public, I'd mark as internal. Or move it into a wrapper/factory.

WendellAdriel (changes requested):

WendellAdriel (approved):
LGTM :muscle:

PR #230, opened 2026-06-01
Extract caching into RegisteredClient

pushpak1300:
The original implementation of named-client caching (#228) worked by injecting a `PrimitiveCache` into `Client` itself via a public `withListCache()` method. That meant `Client` carried cache state even when used inline (not registered), and the `@internal` tag was the only thing stopping users from calling it directly.

Before:

```php
// ClientManager wired it this way
$factory()->withListCache($cache);   // mutates Client directly

// Client held the cache
public function withListCache(?PrimitiveCache $cache): static { ... }
public function flushCache(): void { ... }
```

After:

```php
// ClientManager wraps the Client in a RegisteredClient
new RegisteredClient($factory(), $cache);

// Mcp::client() now returns RegisteredClient
$client = Mcp::client('notion');   // RegisteredClient, not Client
$client->tools();                  // cached here
$client->flushCache();             // owned by RegisteredClient

// Inline clients are unchanged and completely cache-unaware
$client = new Client($transport);
$client->tools();                  // no cache, as before
```

`Client` loses `listCache`, `withListCache()`, and `flushCache()` entirely. The only internal seam left is `fetchToolPayloads()` on `Client`, which `RegisteredClient` calls on a cache miss to get raw payloads for storage. `ListTools` is simplified back to a plain fetch+hydrate with no `PrimitiveCache` dependency.

`PrimitiveCache::flush()` is also tightened to require an explicit `string $kind` argument, and `flushAll()` is added to clear all known kinds at once.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace Laravel\Mcp;
 
 use Illuminate\Support\Collection;
-use Laravel\Mcp\Client\Cache\PrimitiveCache;
 use Laravel\Mcp\Client\Contracts\Transport;
 use Laravel\Mcp\Client\Methods\Ping;
 use Laravel\Mcp\Client\Methods\Tools\CallTool;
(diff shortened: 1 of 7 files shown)


WendellAdriel (approved):
LGTM :fire:

PR #232, opened 2026-06-01
Add OAuth discovery and encrypted auth storage

pushpak1300:
This is the first of a few stacked PRs breaking up the MCP client OAuth work into reviewable pieces. It adds just the protocol/storage foundation, no routes or grant flows yet, so it's easy to review in isolation.

What's here:

- `AuthServerDiscovery` plus the `AuthServerMetadata` / `ProtectedResourceMetadata` value objects for OAuth 2.0 authorization server and protected resource metadata discovery (RFC 8414 / RFC 9728).
- `WwwAuthenticateChallenge` for parsing Bearer `WWW-Authenticate` challenges.
- `TokenSet` to hold and expire access/refresh tokens.
- `EncryptedCacheStore`, a cache wrapper that encrypts payloads at rest (used later to persist tokens).
- `OAuthException` and `DiscoveryException`.

No user-facing API yet, the client integration lands in follow-up PRs.


src/Client/Auth/AuthServerDiscovery.php
```diff
@@ -0,0 +1,243 @@
...
+    {
+        $data = $this->decodeJson($url, $response);
+
+        $authorizationServers = Arr::get($data, 'authorization_servers');
+
+        if (! is_array($authorizationServers) || $authorizationServers === []) {
+            throw new DiscoveryException("Protected resource metadata at [{$url}] is missing [authorization_servers].");
+        }
+
+        return new ProtectedResourceMetadata(
```
WendellAdriel, line 174: I was taking a look at the OAuth RFCs and I think this should validate the returned `resource` value before trusting the metadata. RFC 9728 requires it to exactly match the MCP resource URL, otherwise we may accept metadata for the wrong protected resource and use the wrong authorization server.

src/Client/Auth/AuthServerDiscovery.php
```diff
@@ -0,0 +1,243 @@
...
+
+        $this->assertHttps($tokenEndpoint);
+
+        $authorizationEndpoint = filled($authorizationEndpoint = Arr::get($data, 'authorization_endpoint')) ? (string) $authorizationEndpoint : null;
+
+        if ($authorizationEndpoint !== null) {
+            $this->assertHttps($authorizationEndpoint);
+        }
+
+        return new AuthServerMetadata(
```
WendellAdriel, line 199: Similar to the above regarding RFCs, let's validate the returned `issuer` before creating `AuthServerMetadata`. RFC 8414 requires it to exactly match the issuer being discovered, and skipping that check leaves this open to authorization-server mix-up issues.

WendellAdriel (changes requested):
Just added two comments regarding OAuth RFCs.

PR #233, opened 2026-06-02
Add OAuth layer for MCP web client

pushpak1300:
This PR adds [OAuth authentication](https://modelcontextprotocol.io/specification/draft/basic/authorization) to the MCP web client, so your application can connect to remote MCP servers that sit behind OAuth (GitHub, Notion, your own internal servers, and so on).

The package takes care of the OAuth handshake for you. It handles discovery, PKCE, the `state` and `iss` checks, the authorization code exchange, dynamic client registration, and the refresh and client credentials grants. At the end it hands you back a `TokenSet`. Your application decides where to store that token, and passes it back on each call with `withToken()`.

You stay in control of your routes and your storage. The package never stores the token for you.

## Getting started

First register your client. This usually lives in a service provider, for example `app/Providers/AppServiceProvider.php`:

```php
use Laravel\Mcp\Client;
use Laravel\Mcp\Facades\Mcp;

Mcp::registerClient('github', fn () =>
    Client::web('https://api.githubcopilot.com/mcp')->withOAuth()
);
```

## Connecting a user (Dynamic Client Registration)

This is the most common flow. A user clicks "Connect", gets redirected to the provider to approve access, and comes back to your app with a token. When you leave out the `clientId` and `clientSecret`, the package registers a client on the fly using Dynamic Client Registration.

Register the routes in `routes/web.php`. A single call wires up both the connect route and the callback route for you:

```php
use Laravel\Mcp\Facades\Mcp;
use Laravel\Mcp\Client\OAuth\TokenSet;

Mcp::oAuthRoutesFor('github', function (string $provider, TokenSet $token) {
    auth()->user()->update([
        "mcp_{$provider}_token"   => encrypt($token->accessToken),
        "mcp_{$provider}_refresh" => encrypt($token->refreshToken),
    ]);

    return redirect('/dashboard');
});
```

That gives you two named routes:

- `mcp.oauth.github.connect` at `/mcp/github/connect`, which starts the flow
- `mcp.oauth.github.callback` at `/mcp/oauth/github/callback`, which receives the token

Now link to the connect route from your view:

```blade
<a href="{{ route('mcp.oauth.github.connect') }}">Connect GitHub</a>
```

If you would rather keep your handler in a controller, you may pass it as an array. The method receives the provider name and the token, and resolves through the container so you can type hint anything else you need:

```php
Mcp::oAuthRoutesFor('github', [GitHubConnectionController::class, 'store']);
```

Your handler owns the final redirect, so you decide where the user lands after connecting. You may also choose the middleware applied to both routes:

```php
Mcp::oAuthRoutesFor('github', $handler, middleware: ['web', 'auth']);
```

If the server registered you dynamically and your `clientId` was issued on the fly, the new credentials are available on the token so you can store them for later refreshes:

```php
Mcp::oAuthRoutesFor('github', function (string $provider, TokenSet $token) {
    auth()->user()->update([
        "mcp_{$provider}_token"     => encrypt($token->accessToken),
        "mcp_{$provider}_refresh"   => encrypt($token->refreshToken),
        "mcp_{$provider}_client_id" => encrypt((string) $token->clientId),
        "mcp_{$provider}_client_secret" => encrypt((string) $token->clientSecret),
    ]);

    return redirect('/dashboard');
});
```

## Server to server (client credentials)

For background jobs or service to service access where there is no user to redirect, provide a `clientId` and `clientSecret` and ask for a token directly. No routes are needed here:

```php
use Laravel\Mcp\Facades\Mcp;

Mcp::registerClient('billing', fn () => Client::web('https://billing.test/mcp')->withOAuth(
    clientId: env('BILLING_CLIENT_ID'),
    clientSecret: env('BILLING_CLIENT_SECRET'),
    scope: 'mcp:use',
));

$token = Mcp::client('billing')->oAuthClient()->clientCredentials();

cache()->put('mcp_billing_token', $token->accessToken, $token->expiresAt - time());
```

The package sends the client credentials in whichever way the authorization server advertises (it reads the server metadata and uses `client_secret_basic` or `client_secret_post` accordingly), so there is nothing to configure here.

## Using the token

Pass your stored token on each call with `withToken()`. It accepts a string, or a closure so you can resolve it lazily:

```php
$tools = Mcp::client('billing')
    ->withToken(fn () => cache()->get('mcp_billing_token'))
    ->tools();

$tools = Mcp::client('github')
    ->withToken(auth()->user()->mcp_github_token)
    ->tools();
```

## When the token is missing or expired

The package does not track expiry, so the signal that a user needs to reconnect is the server answering with a `401` or `403`. When that happens the client throws an `AuthorizationRequiredException`, which you can catch to send the user back through the connect route:

```php
use Laravel\Mcp\Client\Exceptions\AuthorizationRequiredException;

try {
    return Mcp::client('github')->withToken($user->mcp_github_token)->tools();
} catch (AuthorizationRequiredException $e) {
    return redirect()->route('mcp.oauth.github.connect', $e->query());
}
```

Passing `$e->query()` forwards the server's challenge details (the resource metadata URL and the scope it asked for) to the connect route, so the reconnect uses exactly what the server requested.

If you are combining tools from several servers and would rather skip one that is not connected yet, pass a default and the listing returns that instead of throwing:

```php
$tools = Mcp::client('github')
    ->withToken(fn () => $user->mcp_github_token)
    ->tools(default: []);
```

## Refreshing a token

Refresh is driven by your application, using the refresh token you stored earlier:

```php
$fresh = Mcp::client('github')->oAuthClient()->refreshCredentials($user->mcp_github_refresh);
```

For a dynamically registered client, pass the credentials you stored at connect time so the refresh authenticates as the same client:

```php
$fresh = Mcp::client('github')->oAuthClient()->refreshCredentials(
    $user->mcp_github_refresh,
    clientId: decrypt($user->mcp_github_client_id),
    clientSecret: decrypt($user->mcp_github_client_secret),
);
```

## Where things live

- The package owns the OAuth handshake, the connect and callback routes, and returning the `TokenSet`.
- Your application owns the client registration (a service provider), the route registration and handler (`routes/web.php` or a controller), and storing the token.

### DIFF
--- a/src/Client/OAuth/AuthServerDiscovery.php
+++ b/src/Client/OAuth/AuthServerDiscovery.php
@@ -15,7 +15,7 @@ class AuthServerDiscovery
 
         $this->requireFetchable($metadataUrl, $resourceUrl);
 
-        $resourceMetadata = $this->fetchResourceMetadata($metadataUrl);
+        $resourceMetadata = $this->fetchResourceMetadata($metadataUrl, explicit: $resourceMetadataUrl !== null);
 
         $this->requireResourceMatches($resourceMetadata, $resourceUrl);
 
@@ -48,7 +48,7 @@ class AuthServerDiscovery
     /**
      * @return array<string, mixed>
      */
-    protected function fetchResourceMetadata(string $metadataUrl): array
+    protected function fetchResourceMetadata(string $metadataUrl, bool $explicit = false): array
     {
         $response = Http::acceptJson()
             ->timeout(5)
@@ -57,12 +57,24 @@ class AuthServerDiscovery
             ->get($metadataUrl);
 
         if (! $response->successful()) {
+            if ($explicit) {
+                throw new OAuthException("Protected resource metadata request to [{$metadataUrl}] failed with status [{$response->status()}].");
+            }
+
             return [];
         }
 
         $data = $response->json();
 
-        return is_array($data) ? $data : [];
+        if (is_array($data)) {
+            return $data;
+        }
+
+        if ($explicit) {
+            throw new OAuthException("Protected resource metadata at [{$metadataUrl}] did not return a valid JSON object.");
+        }
+
+        return [];
     }
 
     /**
--- a/src/Client/OAuth/AuthServerDiscovery.php
+++ b/src/Client/OAuth/AuthServerDiscovery.php
@@ -13,6 +13,8 @@ class AuthServerDiscovery
     {
         $metadataUrl = $resourceMetadataUrl ?? $this->wellKnown($resourceUrl, 'oauth-protected-resource');
 
+        $this->requireFetchable($metadataUrl, $resourceUrl);
+
         $resourceMetadata = $this->fetchResourceMetadata($metadataUrl);
 
         $this->requireResourceMatches($resourceMetadata, $resourceUrl);
@@ -20,6 +22,7 @@ class AuthServerDiscovery
         $issuer = $this->issuerFrom($resourceMetadata) ?? $this->origin($resourceUrl);
 
         $this->requireSecure($issuer);
+        $this->requireNotInternal($issuer, $resourceUrl);
 
         $serverMetadata = $this->fetchMetadata($issuer);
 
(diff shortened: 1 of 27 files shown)


src/Client/OAuth/AuthServerDiscovery.php
```diff
@@ -0,0 +1,234 @@
...
+     */
+    protected function fetchResourceMetadata(string $metadataUrl): array
+    {
+        $response = Http::acceptJson()
+            ->timeout(5)
+            ->connectTimeout(2)
+            ->withOptions(['allow_redirects' => false])
+            ->get($metadataUrl);
+
+        if (! $response->successful()) {
```
WendellAdriel, line 58: This returns an empty metadata array for any failed protected-resource metadata request. That is fine for the default discovery probe, but not when `$resourceMetadataUrl` was explicitly passed to `discover()` from a challenge or the connect route. In this scenario, returning `[]` makes `issuerFrom([])` return `null`, and then `discover()` falls back to `$this->origin($resourceUrl)`. So a bad or unavailable challenge-provided metadata URL can silently authenticate against a different issuer. I think we should distinguish default discovery from explicit discovery and throw when the explicit metadata URL returns non-2xx or invalid JSON.

WendellAdriel (changes requested):

src/Client/OAuth/AuthServerDiscovery.php
pushpak1300: yeah great catch this is valid issue.

WendellAdriel (approved):
LGTM :partying_face:

PR #235, opened 2026-06-03
Remove MCP client tool-list caching layer

pushpak1300:
The MCP client shipped a built-in tool-list cache. It fought the package's thin, app-owns-storage philosophy and added a per-user cache-scoping footgun. It also kept getting more awkward as the OAuth work landed: the API for configuring it grew weirder the more cases it tried to cover. In practice most apps don't need to cache the tools list at all, and the ones that do already have a cache they trust. So this hands caching back to the application. The package just gives you a correct tools list, and you store it however you like.

Before, you configured caching through the package:

```php
Mcp::registerClient('github', fn () => Client::web($url), cacheTtl: 3600, scope: fn () => auth()->id());
$tools = Mcp::client('github')->tools();   // RegisteredClient wrapper, cached for you
Mcp::client('github')->flushCache();
```

After, registration is just a name and a factory, and the tools list is a plain value you can cache directly:

```php
Mcp::registerClient('github', fn () => Client::web($url));

// no cache: just fetch
$tools = Mcp::client('github')->tools();

// app-owned cache: the tools list is plain data, so cache it as-is under your own key
$tools = Cache::remember('mcp:github:tools', 3600, fn () => Mcp::client('github')->tools());
```


src/Client/Primitives/Tool.php
```diff
@@ -64,12 +60,4 @@ public static function from(Client $client, array $payload): self
             meta: $meta,
         );
     }
-
-    /**
-     * @param  array<string, mixed>  $arguments
-     */
-    public function call(array $arguments = []): ToolResult
```
WendellAdriel, line 71: This seems a breaking change on public API, is this expected?

pushpak1300: Yes. we haven;t release this feature yet so its not breaking change.

WendellAdriel (approved):
LGTM :fire:

PR #239, opened 2026-06-04
Make MCP client tools cacheable via recipe serialization

pushpak1300:
Currently you can't cache or queue an MCP client's tool list. A connected client holds a live `Symfony\Component\Process` (stdio) or a session id (http), so the moment you try to stash the result it blows up on serialization:

```php
// throws: cannot serialize a live Process / stale session
Cache::remember('tools', 3600, fn () => Mcp::client('everything')->tools());
```

This PR makes clients serialize a reconstruction recipe instead of the live connection, and reconnect lazily on first use. The tools that come back from `tools()` are now callable too:

```php
$tools = Cache::remember('tools', 3600, fn () => Mcp::client('everything')->tools());

// later, in a queued job / next request, straight from cache:
$tools['add']->call(['a' => 1, 'b' => 2]);
```

Named clients serialize as just their name and re-resolve through the manager on wake, so registered tokens never land in the cache and you always reconnect with a fresh session. Ad-hoc clients (`Client::web(...)->withToken(...)`) serialize their own transport recipe, token included, since there's no registry to re-resolve from.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -4,7 +4,9 @@ declare(strict_types=1);
 
 namespace Laravel\Mcp;
 
+use Illuminate\Container\Container;
 use Illuminate\Support\Collection;
+use Laravel\Mcp\Client\ClientManager;
 use Laravel\Mcp\Client\Contracts\Transport;
 use Laravel\Mcp\Client\Methods\Ping;
 use Laravel\Mcp\Client\Methods\Tools\CallTool;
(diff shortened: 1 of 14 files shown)


src/Client.php
```diff
@@ -98,6 +110,42 @@ public function callTool(string $name, array $arguments = []): ToolResult
...
+
+    /**
+     * @param  array<string, mixed>  $data
+     */
+    public function __unserialize(array $data): void
+    {
+        $this->name = $data['name'] ?? null;
+
+        if ($this->name !== null) {
+            $resolved = Container::getInstance()->make(ClientManager::class)->build($this->name);
```
WendellAdriel, line 137: IDK if this is a valid use case, but if someone creates a child class of this, since this is copying only transport and clientInfo, the child class can silently lose its own state after cache or queue restore. If this is a valid use case, we should add an explicit subclass restore hook/copy contract.

WendellAdriel (changes requested):

src/Client.php
pushpak1300: I don't think anyone will be extending Client. but if anyone did that they will have to override serialise and deserialise so we are fine and it will break only when they are trying to store.

WendellAdriel (approved):
LGTM :muscle:

PR #241, opened 2026-06-08
Fix Mcp alias: correct facade class name

adelf:
Looks like you moved the facade to the new namespace, but forgot about alias configuration.


pushpak1300:
Thanks

PR #245, opened 2026-06-09
Add withHeaders() to the MCP web client for custom request headers

leek:
This adds a `withHeaders()` method to the MCP web client so callers can send custom HTTP headers with each request.

### Why

Today the streamable-HTTP client only supports authentication via `withToken()`, which always emits `Authorization: Bearer {token}`. That doesn't cover MCP servers that require:

- a **raw `Authorization` header** (e.g. an API key sent with no `Bearer` prefix), or
- **additional custom headers** such as a tenant id or an API-version pin.

There's currently no public way to set these, so those servers can't be reached at all. For example, Stedi's healthcare MCP server documents `Authorization: <api-key>` with no scheme prefix — `withToken()` produces `Bearer <api-key>`, which the server rejects.

### Usage

```php
use Laravel\Mcp\Client;

$tools = Client::web('https://mcp.example.com/mcp')
    ->withHeaders([
        'Authorization' => 'Key '.$apiKey,        // sent verbatim, no Bearer prefix
        'X-Tenant' => 'acme',
        'X-Signature' => fn (): string => sign(), // lazily resolved per request
    ])
    ->tools();
```

### Behavior & precedence

- Custom headers are the base set. The transport's protocol-managed headers (`Accept`, `MCP-Session-Id`, `MCP-Protocol-Version`) always take precedence and cannot be clobbered — **case-insensitively**, so a lowercase `accept` or `mcp-session-id` can't slip a duplicate value onto the wire.
- If a bearer token is also set via `withToken()`, its `Authorization: Bearer …` wins over a custom `Authorization` (any casing), so the two auth paths never conflict. When no token is set, a custom `Authorization` header is sent verbatim.
- Header values accept `string` or `Closure(): string`, mirroring `withToken()` — useful for request signing or resolving a secret lazily. Closures are resolved on each request and when serializing the recipe.
- Headers are re-applied when a transport is rebuilt from its `recipe()` — e.g. anonymous clients serialized into queued jobs, and named clients (which the factory rebuilds on each `build()`). As with the bearer token, resolved header values are persisted in the recipe, so treat any secret they carry accordingly (documented on the method).

### No breaking changes

Purely additive: `withHeaders()` is new, and behavior is unchanged when it isn't called. The transport `recipe()` gains an additive `headers` key; recipes serialized before this change (without the key) rebuild safely.

### Tests

Added coverage in `HttpTransportTest`:
- custom header sent; raw `Authorization` without a `Bearer` prefix; bearer-token precedence over a custom `Authorization` (canonical **and** lowercase casing); lowercase `accept` not doubling the protocol header; protocol-managed headers (`mcp-session-id`, `MCP-Protocol-Version`) cannot be injected via `withHeaders`; accumulation across multiple `withHeaders()` calls; closure header values resolved per request and when serializing the recipe; an end-to-end fluent test through `Client::web`.

And in `TransportFactoryTest`: headers round-trip via the recipe, plus a legacy recipe (no `headers` key) rebuilding safely.

The full suite passes (958 tests), with Pint and PHPStan clean.

### DIFF
--- a/src/Client/Transport/HttpTransport.php
+++ b/src/Client/Transport/HttpTransport.php
@@ -23,6 +23,9 @@ class HttpTransport implements Transport
     /** @var string|(Closure(): string)|null */
     protected string|Closure|null $token = null;
 
+    /** @var array<string, string|(Closure(): string)> */
+    protected array $headers = [];
+
     protected ?string $sessionId = null;
 
     protected bool $initialized = false;
@@ -62,6 +65,21 @@ class HttpTransport implements Transport
         $this->token = $token;
     }
 
+    /**
+     * Add custom headers sent with every request.
+     *
+     * Reserved protocol headers (Accept, MCP-Session-Id, MCP-Protocol-Version) and
+     * the bearer token always take precedence and cannot be overridden, regardless
+     * of header-name casing. Values are persisted in the serialized recipe like the
+     * bearer token, so treat any secret they carry accordingly.
+     *
+     * @param  array<string, string|(Closure(): string)>  $headers
+     */
+    public function withHeaders(array $headers): void
+    {
+        $this->headers = array_merge($this->headers, $headers);
+    }
+
     public function url(): string
     {
         return $this->url;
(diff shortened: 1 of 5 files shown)


taylorotwell:

PR #250, opened 2026-06-10
Add withHeaders() to the MCP web client

### LINKED ISSUE #247: Web client can't send a raw Authorization header or custom request headers
The streamable-HTTP web client only exposes `withToken()` for auth, and that always sends `Authorization: Bearer {token}` ([HttpTransport.php#L173](https://github.com/laravel/mcp/blob/main/src/Client/Transport/HttpTransport.php#L173)). There's no public way to send a bare `Authorization` value or any other custom header.

This makes some MCP servers unreachable. Stedi's healthcare MCP server expects `Authorization: <api-key>` with no scheme prefix - `withToken()` turns that into `Bearer <api-key>`, which the server rejects. Same problem for servers that require an additional header like a tenant id or an API-version pin.

I took a shot at this in #245 before it was closed - filing this so the underlying problem is tracked.

pushpak1300:
Currently the streamable-HTTP web client only exposes withToken() for auth, which always sends `Authorization: Bearer {token}`. That leaves out servers that need a raw Authorization header (e.g. Stedi's healthcare server expects `Authorization: <api-key>` with no scheme) or extra headers like a tenant id or API version. There's no way to reach those servers today.

This PR adds withHeaders():

```php
Client::web('https://mcp.example.com/mcp')
    ->withHeaders([
        'Authorization' => $apiKey,   // sent verbatim, no Bearer prefix
        'X-Tenant-Id' => 'acme',
    ])
    ->tools();
```

Custom headers take precedence over the transport's defaults and override them case-insensitively, so a lowercase `accept` replaces the default rather than sending a duplicate header on the wire. Headers round-trip through the transport recipe, so named/serialized clients keep them.

Fixes #247

### DIFF
--- a/src/Client/Transport/HttpTransport.php
+++ b/src/Client/Transport/HttpTransport.php
@@ -29,6 +29,9 @@ class HttpTransport implements Transport
 
     protected float $timeoutSeconds = 30.0;
 
+    /** @var array<string, string> */
+    protected array $customHeaders = [];
+
     /** @var array<int, string> */
     protected array $queue = [];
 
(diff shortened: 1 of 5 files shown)


pushpak1300:
Credits to https://github.com/laravel/mcp/pull/245/ @leek

PR #254, opened 2026-06-25
Allow OAuth client registration without a client name

### LINKED ISSUE #253: OAuth DCR rejects registrations without `client_name`, contrary to RFC 7591 (breaks Claude connector)
## Summary

`OAuthRegisterController` makes the OAuth Dynamic Client Registration (DCR) request **reject** any registration that does not include `client_name` (or the legacy `name`). However, [RFC 7591 §2](https://datatracker.ietf.org/doc/html/rfc7591#section-2) lists `client_name` as an **OPTIONAL** client metadata field. As a result, spec-compliant MCP clients that omit `client_name` cannot register and receive `400 invalid_client_metadata`.

This breaks real-world clients — notably **Claude's remote MCP connector** (claude.ai), whose DCR request is rejected with `400 invalid_client_metadata`, surfacing to users as *"Couldn't register with … sign-in service"*.

## Where

`src/Server/Http/Controllers/OAuthRegisterController.php`:

```php
'client_name' => ['nullable', 'string', 'min:1', 'max:255', 'required_without:name'],
'name'        => ['nullable', 'string', 'min:1', 'max:255', 'required_without:client_name'],
```

The `required_without` rules force one of the two to be present. This requirement was added intentionally in #177, but it is stricter than RFC 7591, which the MCP authorization spec builds on for DCR.

## Reproduction

```bash
# Missing client_name -> 400 (RFC 7591 says this is a valid request)
curl -s -X POST https://<host>/<oauth-prefix>/register \
  -d '{"redirect_uris":["https://example.com/callback"],"token_endpoint_auth_method":"none","grant_types":["authorization_code","refresh_token"],"response_types":["code"]}'
# => 400 {"error":"invalid_client_metadata","error_description":"The client name field is required when name is not present."}

# Same request WITH client_name -> 201
curl -s -X POST https://<host>/<oauth-prefix>/register \
  -d '{"client_name":"X","redirect_uris":["https://example.com/callback"],"token_endpoint_auth_method":"none","grant_types":["authorization_code","refresh_token"],"response_types":["code"]}'
# => 201 { "client_id": "...", ... }
```

Observed in production against Claude's connector backend (its DCR `POST .../oauth/register` is rejected `400`, and the identical `400 invalid_client_metadata` reproduces by omitting `client_name`).

## Expected

Per RFC 7591 §2, `client_name` is optional. The registration endpoint should accept requests without it rather than returning `400`.

## Suggested fix

Make `client_name`/`name` optional, and when both are absent, fall back to a sensible default instead of rejecting — e.g. derive a name from the registered `redirect_uris` host, or store `null`. The existing `client_name` > `name` precedence can be preserved. For example:

```php
'client_name' => ['nullable', 'string', 'min:1', 'max:255'],
'name'        => ['nullable', 'string', 'min:1', 'max:255'],
// ...
$name = $validated['client_name']
    ?? $validated['name']
    ?? (parse_url($validated['redirect_uris'][0] ?? '', PHP_URL_HOST) ?: 'MCP Client');
```

## Environment

- `laravel/mcp` v0.7.2
- Affected client: Claude remote MCP connector (claude.ai)

Happy to open a PR for this if the direction is acceptable.

lazerg:
RFC 7591 §2 lists `client_name` as an OPTIONAL field for Dynamic Client Registration, but `OAuthRegisterController` rejects any registration that omits both `client_name` and the legacy `name`, returning `400 invalid_client_metadata`. This blocks spec-compliant clients — notably Claude's remote MCP connector, whose registration request leaves out `client_name` and currently fails with *"Couldn't register with … sign-in service"*.

This PR makes `client_name`/`name` optional. When neither is supplied, the client name falls back to the host of the first redirect URI (e.g. `example.com`), and to `MCP Client` when no host can be derived. The existing `client_name` > `name` precedence is unchanged, so requests that already send a name behave exactly as before.

Tests are updated to cover a registration request without a name and the fallback behaviour.

Fixes #253

### DIFF
--- a/src/Server/Http/Controllers/OAuthRegisterController.php
+++ b/src/Server/Http/Controllers/OAuthRegisterController.php
@@ -75,15 +75,9 @@ class OAuthRegisterController
             'Laravel\Passport\ClientRepository'
         );
 
-        $redirectUris = (array) $validated['redirect_uris'];
-
-        $name = $validated['client_name']
-            ?? $validated['name']
-            ?? (parse_url((string) ($redirectUris[0] ?? ''), PHP_URL_HOST) ?: 'MCP Client');
-
         $client = $clients->createAuthorizationCodeGrantClient(
-            name: $name,
-            redirectUris: $redirectUris,
+            name: $this->resolveClientName($validated),
+            redirectUris: $validated['redirect_uris'],
             confidential: false,
             enableDeviceFlow: false,
         );
@@ -96,6 +90,18 @@ class OAuthRegisterController
             'scope' => 'mcp:use',
             'token_endpoint_auth_method' => 'none',
         ], 201);
+    }
+
+    /**
+     * Resolve the client name, falling back to the redirect host or a default.
+     *
+     * @param  array<string, mixed>  $validated
+     */
+    protected function resolveClientName(array $validated): string
+    {
+        return $validated['client_name']
+            ?? $validated['name']
+            ?? (parse_url((string) ($validated['redirect_uris'][0] ?? ''), PHP_URL_HOST) ?: 'MCP Client');
     }
 
     protected function isValidRedirectUri(string $value): bool
--- a/src/Server/Http/Controllers/OAuthRegisterController.php
+++ b/src/Server/Http/Controllers/OAuthRegisterController.php
@@ -21,8 +21,8 @@ class OAuthRegisterController
     public function __invoke(Request $request): JsonResponse
     {
         $validator = Validator::make($request->all(), [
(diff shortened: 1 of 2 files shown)


lazerg:
cc @pushpak1300 — since you built the OAuth/DCR layer in #233, this is likely your area. Small change to make `client_name` optional per RFC 7591 §2, which lets Claude's remote connector register (its DCR request omits `client_name`). Tests, Pint, PHPStan (level 8) and Rector all pass locally; CI is just waiting on the first-time-contributor approval. Thanks for taking a look!

taylorotwell:

pushpak1300:
Thanks.

PR #261, opened 2026-07-07
Fix schema return type in tool stub

axlon:
The stub for tools currently always triggers PHPStan errors on the schema method. This is because its annotated as if it will return `array<string, \Illuminate\Contracts\JsonSchema\JsonSchema>`, but it should in fact return `array<string, \Illuminate\JsonSchema\Types\Type>` (which does not implement the `JsonSchema` contract).

The [docs](https://laravel.com/docs/13.x/mcp#tools) also provide an example where a `Type` return is used

### DIFF
--- a/stubs/mcp-tool.stub
+++ b/stubs/mcp-tool.stub
@@ -24,7 +24,7 @@ class {{ class }} extends Tool
     /**
      * Get the tool's input schema.
      *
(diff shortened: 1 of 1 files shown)


axlon:
Failures seem to have nothing to do with this PR

PR #266, opened 2026-07-12
Add read() to the MCP client resource primitive

dfinchenko:
The client `Tool` primitive is bound to its `Client` and can invoke itself with `->call()`, introduced alongside tool listing in #226. When prompt and resource support landed later (#242, #243), those primitives were list-only, they are never handed the `Client`, so they cannot act on themselves.

That leaves the client API asymmetric. A listed tool reads fluently:

```php
$client->tools()['add']->call(['a' => 1, 'b' => 2]);
```

…but a listed resource has to hand its own URI back to the client to be read:

```php
$resource = $client->resources()['file://readme'];

$client->readResource($resource->uri); // there is no $resource->read()
```

This PR completes the pattern for resources by adding `Resource::read()`, mirroring `Tool::call()` exactly, the primitive is bound to the client during hydration, `read()` delegates to `Client::readResource()`, and an unbound resource throws the same guard exception as an unbound tool:

```php
$resource = $client->resources()['file://readme'];

$resource->read(); // ResourceReadResult
```
### Notes
- Same shape as the existing `Tool::call()` binding: `ListResources` now threads the `Client` through `nextPage()` and into `Resource::from()`, and `Client::resources()` passes `client: $this`.
- No behavioral change to the existing `resources()` / `readResource()` paths, this is purely additive.
- The matching `Prompt::get()` can follow as a separate PR to keep this one focused (mirroring how prompt and resource support were split across #242 and #243).

Tests cover both the bound case (reads through the client and sends `resources/read` with the correct `uri`) and the unbound case (throws until bound).

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -163,7 +163,7 @@ class Client
     public function resources(?int $limit = null, ?iterable $default = null): Collection
     {
         try {
-            return (new ListResources(limit: $limit))->handle($this->protocol);
+            return (new ListResources(client: $this, limit: $limit))->handle($this->protocol);
         } catch (AuthorizationRequiredException $authorizationRequiredException) {
             if ($default === null) {
                 throw $authorizationRequiredException;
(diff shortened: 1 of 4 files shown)


pushpak1300:
This is breaking change.

pushpak1300:
Thanks for your pull request!

Unfortunately, I'm going to delay merging this code for now as it introduces a breaking change to the package.

PR #269, opened 2026-07-15
Serialize structured content before asserting

chrisvanlier2005:
Currently, the actual structuredContent sent in the response is not serialized when testing. Serialization only happens in `\Laravel\Mcp\Server` which is skipped when testing, which means the structured content being asserted is not the same as what is actually sent over the wire to the client.
Because of this, assertions produce a lot of unexpected behavior, as AssertableJson expects serialized/deserialized JSON.

For example:
```php
public function handle(): ResponseFactory                                                                                                                  
{                                                                                                                                                          
    return Response::structured([                                                                                                                          
        'status' => EmployeeStatus::ToDo                                                                                       
    ]);                                                                                                                                                    
}                                                                                                                                                          
                                                                                                                                                             
$response->assertStructuredContent(                                                                                                                        
    fn (AssertableJson $json): AssertableJson => $json                                                                                                     
        ->where('status', EmployeeStatus::Todo)                                                                                     
        ->etc()                                                                                                                                            
);
```
Currently does not work:
> Property [status] does not match the expected value.
Failed asserting that \EmployeeStatus Enum #5278 (ToDo, 'to_do') is identical to 'to_do'.

In this case "expected" enums are converted to strings for expectations. Which is correct, but the content in `AssertableJson` was never actually serialized to json. So the assertion fails.

When classes are used in the response, `where` always fails, as its checked by reference instead of value like the api `->assertJson`

### DIFF
--- a/src/Server/Testing/TestResponse.php
+++ b/src/Server/Testing/TestResponse.php
@@ -9,6 +9,7 @@ use Illuminate\Container\Container;
 use Illuminate\Contracts\Auth\Authenticatable;
 use Illuminate\Support\Traits\Conditionable;
 use Illuminate\Support\Traits\Macroable;
+use Illuminate\Testing\AssertableJsonString;
 use Illuminate\Testing\Fluent\AssertableJson;
 use Laravel\Mcp\Server\Primitive;
 use Laravel\Mcp\Server\Prompt;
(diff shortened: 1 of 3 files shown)


pushpak1300:
Added some test for fix. thanks.

pushpak1300 (approved):

PR #270, opened 2026-07-15
Fix DCR scope persistence for Passport clients

### LINKED ISSUE #258: Grant `mcp:use` scope on oauth clients created via DCR
The client registration controller doesn't set any scopes on the client, but does report `mcp:use` as having been granted to the client. This package also reports `mcp:use` as a scope through protected resource metadata ([source](https://github.com/laravel/mcp/blob/478df1b7ac3b264e5eb11f72fac75f8ea855101e/src/Server/Registrar.php#L176)).

Passport supports limiting what scopes a client may request, it does this automatically when a `scopes` column exists ([source](https://github.com/laravel/passport/blob/87608729a291dfcae6027887de5effec16ebf979/src/Client.php#L212-L215)).

When an application uses this security feature, clients created through dynamic registration never receive the advertised scope and thus cannot request it on tokens.

Gujiassh:
## Summary

- Grant dynamically registered Passport clients the advertised `mcp:use` scope when the client is actually scope-restricted (a non-NULL `scopes` value, e.g. a column defaulting to `'[]'`)
- Merge `mcp:use` into any scopes granted at creation instead of overwriting them; skip the write when the scope is already present
- Leave unrestricted clients untouched: a NULL `scopes` value (or no `scopes` column) is unrestricted per Passport's `Client::hasScope`, and stays that way
- Wrap client creation + scope grant in a transaction and return an RFC 7591 `server_error` JSON response on failure instead of an unhandled 500
- Centralize the `mcp:use` literal as `Registrar::OAUTH_SCOPE`

Fixes #258.

## Behavior notes

- Existing OAuth clients are not migrated or modified; no schema change is included
- The transaction uses the default database connection; installations pointing `passport.connection` elsewhere keep the fix but not the atomicity guarantee

## Testing

- Database-backed registration tests cover: restricted-by-default schema (`'[]'`), merging with creation-time scopes, no duplicate grant, NULL stays unrestricted, and no `scopes` column
- `vendor/bin/pest`, `composer test:types`, `vendor/bin/pint --test`, `vendor/bin/rector --dry-run` all pass

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -48,7 +48,7 @@ jobs:
         uses: shivammathur/setup-php@f3e473d116dcccaddc5834248c87452386958240 # v2
         with:
           php-version: ${{ matrix.php }}
-          extensions: dom, curl, libxml, mbstring, zip, fileinfo
+          extensions: dom, curl, libxml, mbstring, zip, fileinfo, pdo_sqlite, sqlite3
           ini-values: error_reporting=E_ALL
           tools: composer:v2
           coverage: xdebug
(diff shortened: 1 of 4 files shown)


pushpak1300:
Thanks

pushpak1300 (approved):

PR #271, opened 2026-07-16
Fix multiline SSE data event parsing

Gujiassh:
## Summary

- buffer SSE `data` fields until the event's blank-line boundary
- join multiple `data` values with newlines as required by the SSE event format
- dispatch a final event when the stream ends without a trailing blank line

Before this change, each `data:` line was queued as a separate JSON-RPC message. A valid response split across multiple data fields therefore produced malformed fragments instead of one response.

## Tests

- `vendor/bin/pest tests/Unit/Client/Transport/HttpTransportTest.php` (32 tests, 75 assertions)
- `vendor/bin/pest --ci` (1007 tests, 2283 assertions)
- `vendor/bin/phpstan analyse --no-progress`
- `vendor/bin/pint --test src/Client/Transport/HttpTransport.php tests/Unit/Client/Transport/HttpTransportTest.php`
- `vendor/bin/rector process --dry-run src/Client/Transport/HttpTransport.php tests/Unit/Client/Transport/HttpTransportTest.php`

### DIFF
--- a/src/Client/Transport/HttpTransport.php
+++ b/src/Client/Transport/HttpTransport.php
@@ -218,14 +218,28 @@ class HttpTransport implements Transport
     protected function readSseStream(ClientResponse $response): void
     {
         $stream = $response->toPsrResponse()->getBody();
+        $dataLines = [];
 
         while (! $stream->eof()) {
-            $line = trim($this->readLine($stream));
+            $line = rtrim($this->readLine($stream), "\r\n");
 
-            if (Str::startsWith($line, 'data:')) {
-                $this->queueSseEvent(trim(Str::after($line, 'data:')));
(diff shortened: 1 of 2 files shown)


taylorotwell:
Is this linked to an actual issue or is theoretical?

Gujiassh:
There is no linked user issue; I found this during a transport code audit.

I reproduced it on current `main` with a valid SSE event split across two `data` fields. The current implementation queues each field separately, so the first JSON fragment decodes to `null` instead of producing one JSON-RPC response at the event boundary.

I do not have a report from a real integration, so treating this as theoretical is fair. I will leave it closed and avoid audit-only fixes here without a concrete issue or integration report.

PR #279, opened 2026-07-31
Sort the WWW-Authenticate challenge outside authentication

### LINKED ISSUE #278: AddWwwAuthenticateHeader never runs when the MCP route is protected by auth middleware
### Laravel Package Version

v0.9.1

### Laravel Version

13.15.0

### PHP Version

8.5.8

### Database Driver & Version

PostgreSQL 18.0

### Description

`AddWwwAuthenticateHeader` never runs on an OAuth-protected MCP server, so the `401` carries no `WWW-Authenticate` challenge and a spec-following client cannot discover the authorization server.

`Registrar::web()` attaches the middleware at route level:

```php
$route = Router::post($route, ...)->middleware([
    ReorderJsonAccept::class,
    AddWwwAuthenticateHeader::class,
]);
```

The middleware decorates a **returned** 401:

```php
$response = $next($request);
if ($response->getStatusCode() !== 401) {
    return $response;
}
```

so it only works from outside whatever produces one. But protecting the route means adding an auth middleware, and Laravel sorts `Illuminate\Auth\Middleware\Authenticate` to the front of the stack because it implements `Illuminate\Contracts\Auth\Middleware\AuthenticatesRequests`, which sits in the framework's `$middlewarePriority` list. `AddWwwAuthenticateHeader` is not in that list and therefore cannot be sorted ahead of it.

For this route:

```php
Mcp::web('/mcp', MyServer::class)->middleware(['auth:api', 'my.middleware']);
```

the resolved order is:

```
0. Illuminate\Auth\Middleware\Authenticate:api
1. Illuminate\Routing\Middleware\SubstituteBindings
2. Laravel\Mcp\Server\Middleware\ReorderJsonAccept
3. Laravel\Mcp\Server\Middleware\AddWwwAuthenticateHeader   <-- inside auth
```

Auth produces the 401 at the outermost layer, so the response never travels back out through the decorator.

The effect is that the middleware works in configurations that do not need it, and silently does nothing in the OAuth-protected case it exists for. There is no error, and `/.well-known/oauth-protected-resource/{path}` keeps serving correctly, so nothing indicates the challenge is missing. Clients that hardcode the conventional metadata address still connect, which hides it further; only a client that relies on the challenge for discovery fails, and it fails with no diagnostic.

RFC 9728 §5.1 and the MCP authorization spec both have the client read `WWW-Authenticate` off the 401 and follow `resource_metadata`.

Confirmed on v0.9.1: `Registrar::web()` still attaches the middleware at route level only, and `AddWwwAuthenticateHeader` is unchanged from v0.8.2.

Possible fixes, smallest first:

1. Have the service provider push `AddWwwAuthenticateHeader` onto the priority list ahead of `AuthenticatesRequests`.
2. Catch `AuthenticationException` rather than inspecting a returned response.
3. Register it as group middleware wrapping the MCP routes, outside route-level auth.

Two closed issues are consistent with this symptom, though I have not verified either shares the cause: #210 (Claude can't connect using OAuth) and #139 (OAuth not working with Gemini CLI).

Workaround, for anyone hitting this before it is fixed upstream, in `bootstrap/app.php`:

```php
$middleware->prependToPriorityList(
    before: \Illuminate\Contracts\Auth\Middleware\AuthenticatesRequests::class,
    prepend: \Laravel\Mcp\Server\Middleware\AddWwwAuthenticateHeader::class,
);
```

With that in place the resolved order becomes `AddWwwAuthenticateHeader` at 0, `Authenticate:api` at 1, and the challenge is emitted correctly.

### Steps To Reproduce

1. Register an OAuth-protected MCP server:

```php
// routes/ai.php
Mcp::oauthRoutes();
Mcp::web('/mcp', MyServer::class)->middleware(['auth:api']);
```

2. Send an unauthenticated request:

```bash
curl -i -X POST https://example.test/mcp \
  -H 'Accept: application/json, text/event-stream' \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-06-18","capabilities":{},"clientInfo":{"name":"probe","version":"0"}}}'
```

**Expected:** `401` with

```
WWW-Authenticate: Bearer realm="mcp", resource_metadata="https://example.test/.well-known/oauth-protected-resource/mcp"
```

**Actual:** `401` with no `WWW-Authenticate` header.

3. To see why, inspect the resolved order:

```php
foreach (app('router')->getRoutes() as $r) {
    if ($r->uri() === 'mcp' && in_array('POST', $r->methods())) {
        dump(app('router')->gatherRouteMiddleware($r));
    }
}
```

`AddWwwAuthenticateHeader` appears after `Authenticate`.

jspekken:
Fixes #278

`AddWwwAuthenticateHeader` decorates a **returned** 401:

```php
$response = $next($request);
if ($response->getStatusCode() !== 401) {
    return $response;
}
```

so it only works from outside whatever produces one. But protecting an MCP route means adding an auth middleware, and Laravel sorts `Authenticate` to the front of the stack because it implements `AuthenticatesRequests`, which is in the framework's middleware priority list. `AddWwwAuthenticateHeader` is not in that list, so it cannot be sorted ahead of auth.

For `Mcp::web('/mcp', MyServer::class)->middleware(['auth:api'])` the resolved order was:

```
0. Illuminate\Auth\Middleware\Authenticate:api
1. Illuminate\Routing\Middleware\SubstituteBindings
2. Laravel\Mcp\Server\Middleware\ReorderJsonAccept
3. Laravel\Mcp\Server\Middleware\AddWwwAuthenticateHeader   <-- inside auth
```

Auth produced the 401 at the outermost layer, so the response never travelled back out through the decorator, and the challenge was silently absent on exactly the OAuth-protected servers the middleware exists for.

It is quiet because nothing errors and `/.well-known/oauth-protected-resource/{path}` keeps serving correctly, so the setup looks healthy. Clients that hardcode the conventional metadata address still connect. Only a client that relies on `WWW-Authenticate` for discovery fails, per RFC 9728 §5.1 and the MCP authorization spec, and it fails with no diagnostic.

## The change

Registers the middleware in the HTTP kernel's priority list from `McpServiceProvider::boot()`, guarded so it is a no-op when the HTTP kernel is not bound or is not the concrete `Illuminate\Foundation\Http\Kernel` (console, custom kernels). Resolved order after:

```
0. Laravel\Mcp\Server\Middleware\AddWwwAuthenticateHeader   <-- outside auth
1. Illuminate\Auth\Middleware\Authenticate:api
2. Illuminate\Routing\Middleware\SubstituteBindings
3. Laravel\Mcp\Server\Middleware\ReorderJsonAccept
```

This only reorders middleware already present on a route, so routes that do not use `AddWwwAuthenticateHeader` are unaffected.

## Tests

Added a test that registers a web server with `auth` and asserts the challenge sorts ahead of the auth middleware. It is mutation-checked: with the provider change reverted and the test kept, it fails on

```
AddWwwAuthenticateHeader must sort before the auth middleware, or it never
sees the 401 it exists to decorate.
```

Full suite 1028 passed (2333 assertions), Pint clean, PHPStan no errors.

## Notes

I went with the priority registration as the smallest change. Two alternatives, in case you prefer either: catching `AuthenticationException` in the middleware rather than inspecting a returned response, or applying it as group middleware around the MCP routes. Happy to rework it.

Found on v0.9.1 against a production OAuth-protected server; `Registrar::web()` and the middleware are unchanged from v0.8.2 in the relevant respects.

### DIFF
--- a/src/Server/McpServiceProvider.php
+++ b/src/Server/McpServiceProvider.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace Laravel\Mcp\Server;
 
+use Illuminate\Contracts\Http\Kernel as HttpKernelContract;
+use Illuminate\Foundation\Http\Kernel as HttpKernel;
 use Illuminate\Support\Facades\Route;
 use Illuminate\Support\ServiceProvider;
 use Laravel\Mcp\Client\ClientManager;
@@ -15,6 +17,7 @@ use Laravel\Mcp\Console\Commands\MakeServerCommand;
 use Laravel\Mcp\Console\Commands\MakeToolCommand;
 use Laravel\Mcp\Console\Commands\StartCommand;
 use Laravel\Mcp\Request;
+use Laravel\Mcp\Server\Middleware\AddWwwAuthenticateHeader;
 
 class McpServiceProvider extends ServiceProvider
 {
@@ -31,6 +34,7 @@ class McpServiceProvider extends ServiceProvider
 
     public function boot(): void
     {
+        $this->registerMiddlewarePriority();
         $this->registerMcpScope();
         $this->registerRoutes();
         $this->registerContainerCallbacks();
@@ -66,6 +70,36 @@ class McpServiceProvider extends ServiceProvider
         $this->publishes([
             __DIR__.'/../../config/mcp.php' => config_path('mcp.php'),
         ], 'mcp-config');
+    }
+
+    /**
+     * Ensure the WWW-Authenticate challenge is applied outside authentication.
+     *
+     * AddWwwAuthenticateHeader decorates a returned 401, so it only works from
+     * outside whatever produces one. Protecting an MCP route means adding an
+     * auth middleware, and Laravel sorts Authenticate to the front of the stack
+     * because it implements AuthenticatesRequests, which is in the framework's
+     * middleware priority list. This middleware is not in that list, so it
+     * cannot be sorted ahead of auth and ends up inside it, never seeing the
+     * rejection.
+     *
+     * The result is that the header is silently absent on exactly the
+     * OAuth-protected servers it exists for, with no error and a
+     * correctly-served metadata document to make it look fine.
+     */
+    protected function registerMiddlewarePriority(): void
+    {
+        if (! $this->app->bound(HttpKernelContract::class)) {
+            return;
+        }
+
+        $kernel = $this->app->make(HttpKernelContract::class);
+
+        if (! $kernel instanceof HttpKernel) {
+            return;
+        }
+
+        $kernel->prependToMiddlewarePriority(AddWwwAuthenticateHeader::class);
     }
 
     protected function registerRoutes(): void
--- a/tests/Unit/Server/McpServiceProviderTest.php
+++ b/tests/Unit/Server/McpServiceProviderTest.php
@@ -2,8 +2,12 @@
 
 declare(strict_types=1);
 
+use Illuminate\Auth\Middleware\Authenticate;
 use Laravel\Mcp\Server\McpServiceProvider;
+use Laravel\Mcp\Server\Middleware\AddWwwAuthenticateHeader;
+use Laravel\Mcp\Server\Registrar;
 use Laravel\Passport\Passport;
+use Tests\Fixtures\ExampleServer;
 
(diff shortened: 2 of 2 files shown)


taylorotwell:

PR #281, opened 2026-08-05
Fix mcp:inspector passing mcp:start {handle} as one argument instead of two

johankrijt:
`InspectorCommand` builds `mcp:start {$handle}` as a single array element for the `Process` class. Since Node spawns child processes without a shell, this arrives at artisan as one literal argument, so it tries to run a command named `mcp:start calculations` instead of `mcp:start` + `calculations`, and fails in the background with:

```
Command "mcp:start calculations" is not defined.
```

This breaks mcp:inspector for every local (stdio) MCP server, since the inspector never completes its stdio handshake.

Fix: split it into two array elements: `mcp:start` and `$handle`.

No breaking changes: only affects argument construction for this one Process call; matches how the command already works when run manually.

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -3,7 +3,6 @@
 declare(strict_types=1);
 
 use Rector\CodingStyle\Rector\ClassLike\NewlineBetweenClassLikeStmtsRector;
-use Rector\CodingStyle\Rector\Encapsed\EncapsedStringsToSprintfRector;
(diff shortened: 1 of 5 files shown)


pushpak1300:
Thanks

pushpak1300 (approved):

johankrijt:
My bad!

PR #275, opened 2026-07-25
Add dual-era support for the MCP 2026-07-28 stateless revision

JordanDalton:
## Summary

Implements the [MCP 2026-07-28 revision](https://modelcontextprotocol.io/specification/draft/changelog) (release candidate; final spec publishes July 28) while retaining full compatibility with all previously supported revisions. Servers and clients are now **dual-era**: they speak the new stateless, per-request-metadata protocol with modern peers and fall back to the `initialize` handshake with legacy peers, per the spec's [backward-compatibility rules](https://modelcontextprotocol.io/specification/draft/basic/versioning).

Draft until the final spec text lands — the protocol version string and field names follow the current draft exactly and will be re-verified against the published revision.

### Server

- Serves stateless requests that declare `io.modelcontextprotocol/protocolVersion` in `_meta` — no handshake, no session; `initialize` continues to work for legacy clients and now never negotiates a modern revision (SEP-2575)
- Implements the required `server/discover` method
- Decorates modern results with the required `resultType: "complete"` field and `io.modelcontextprotocol/serverInfo` result `_meta`
- Adds `ttlMs`/`cacheScope` (`CacheableResult`, SEP-2549) to the five cacheable methods plus `server/discover`, configurable per server via `$cacheTtlMs`/`$cacheScope`
- Validates the `MCP-Protocol-Version`, `Mcp-Method`, and `Mcp-Name` request metadata headers against the body (SEP-2243), including the Base64 sentinel encoding, rejecting mismatches with `HeaderMismatch` (`-32020`) and HTTP 400
- Maps modern protocol errors to the spec's HTTP statuses: unknown method → 404, `UnsupportedProtocolVersion` (`-32022`) / header validation → 400
- Missing-resource errors use `-32602` for modern requests (spec alignment), `-32002` for legacy

### Client

- Probes with `server/discover` first and falls back to the legacy handshake on any non-modern error, caching the server's era so reconnects skip the probe
- Stamps `protocolVersion`/`clientInfo`/`clientCapabilities` into `_meta` on every modern request and sends the required request metadata headers
- Surfaces modern JSON-RPC error bodies carried on HTTP 4xx responses (enables version-negotiation retry per spec)
- Treats results without `resultType` as complete; rejects `input_required` (MRTR is not yet supported, and the client advertises no capabilities that would allow it)

### New error/meta plumbing

[...]

PR #287, opened 2026-08-07
Teach the client both protocol eras

pushpak1300:
The client only knew the `initialize` handshake, so it could not talk to a `2026-07-28` server at all, including the one this package now ships.

It now probes `server/discover` on connect and picks an era from the answer:

```php
Mcp::client('everything')->tools();
// server/discover succeeds  -> modern, no handshake, _meta on every request
// server/discover -32601    -> initialize, initialized notification, session ids
```

Pin it when you already know:

```php
Mcp::client('everything')->withEra(ProtocolEra::MODERN)->tools();
```

In modern mode each request carries `io.modelcontextprotocol/protocolVersion`, `clientCapabilities`, and `clientInfo` in `_meta`, plus the `Mcp-Method` and `Mcp-Name` headers #286 validates, with the Base64 sentinel applied to names that are not header safe. `2026-07-28` is never sent through the legacy handshake, and the detected era survives a reconnect.

> [!WARNING]
> BC break. `auto` is the default, so a client that used to open with `initialize` now sends one `server/discover` first. Requests to an MCP endpoint that answers a modern probe with something other than a JSON-RPC error will surface that error instead of falling back. Pass `withEra(ProtocolEra::LEGACY)` to keep the old single-handshake behavior exactly.

`auto` is the default rather than legacy (which is what the TypeScript SDK picks) because this package's own server serves only `2026-07-28`, so a legacy default would fail against it out of the box. The fallback is deliberately narrow: a JSON-RPC error that is not one of the modern protocol codes, or an HTTP status the endpoint rejects the probe with. A timeout or a dead connection propagates rather than being read as "legacy". Tests cover both eras over the fake transport and over faked HTTP, including the sentinel round trip and reading a modern error out of a `400`.

### Spec

- [Versioning: backward compatibility](https://modelcontextprotocol.io/specification/2026-07-28/basic/versioning#backward-compatibility-with-initialization-based-versions)
- [Streamable HTTP: backward compatibility](https://modelcontextprotocol.io/specification/2026-07-28/basic/transports/streamable-http#backward-compatibility)


pushpak1300:
Superseded. The stack is being reordered so every server change lands before the client changes, and the layers are split smaller. Reopened as a new PR at its new position in the stack.

PR #288, opened 2026-08-07
Add the multi round-trip request pattern

pushpak1300:
A server that needs something from the user, an elicitation, a sampling call, a roots listing, used to send its own JSON-RPC request back down the stream. `2026-07-28` removes that. The server returns instead, says what it needs, and the client asks again.

A tool that needs a name:

```php
public function handle(Request $request): Response|InputRequired
{
    if ($request->inputResponses() === []) {
        return InputRequired::make([
            'who' => [
                'method' => 'elicitation/create',
                'params' => ['mode' => 'form', 'message' => 'Who are you?', 'requestedSchema' => $schema],
            ],
        ], requestState: $this->sign($request->all()));
    }

    return Response::text('Hello, '.$request->inputResponses()['who']['content']['name']);
}
```

The client fills it in and retries on its own:

```php
Mcp::client('everything')
    ->onElicitation(fn (array $params) => ['action' => 'accept', 'content' => ['name' => 'John']])
    ->callTool('say-hi');
```

The retry carries a new JSON-RPC id, the `inputResponses` map, and the exact `requestState` the server sent, or no `requestState` at all when the server sent none. Registering a handler is what declares the matching client capability in `_meta`, and the server refuses to ask for a capability the client did not declare, with `-32021`. `InputRequired` is only wired into `tools/call`, `prompts/get`, and `resources/read`.

> [!WARNING]
> BC break for anyone relying on server-initiated requests. A `2026-07-28` server must not send `elicitation/create`, `sampling/createMessage`, or `roots/list` as its own JSON-RPC request, so that path is gone. Use `InputRequired` instead.

`requestState` is deliberately an opaque string the package never inspects: signing or encrypting it is the server author's job, and the spec requires that of anything where tampering matters. The client stops after five input rounds for the same request rather than looping forever. Tests cover the round trip in both directions, the capability refusal, the state echo, and the round limit.

### Spec

- [Multi Round-Trip Requests](https://modelcontextprotocol.io/specification/2026-07-28/basic/patterns/mrtr)
- [SEP-2322](https://modelcontextprotocol.io/seps/2322-MRTR)

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -40,9 +40,6 @@ class Client
 
     protected ProtocolEra $era = ProtocolEra::AUTO;
 
-    /** @var array<string, Closure(array<string, mixed>): array<string, mixed>> */
-    protected array $inputHandlers = [];
-
     public function __construct(
         protected Transport $transport,
         public ?Implementation $clientInfo = null,
@@ -157,8 +154,6 @@ class Client
      */
     protected function onInput(string $method, Closure $handler): static
     {
-        $this->inputHandlers[$method] = $handler;
-
         $this->protocol->onInput($method, $handler);
 
         return $this;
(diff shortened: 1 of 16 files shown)


pushpak1300:
Dropping the multi round-trip request pattern for now — elicitation, sampling, and roots are all new surface, and the goal is to land the rest of the 2026-07-28 work first. The stack has been rebased onto #287.

PR #292, opened 2026-08-07
Document the migration to MCP 2026-07-28

pushpak1300:
Nine protocol changes land together with no single place that tells an upgrader what to do. This PR adds `UPGRADE.md`, with one section per breaking change and its replacement.

```php
// before
{
    return Response::text($request->sessionId());
}

// after: no sessions, pass state as a tool argument
{
    return Response::text($request->get('basket_id'));
}
```

It covers the removed handshake and `SessionInitialized`, required request `_meta`, `server/discover`, removed session APIs, mirrored HTTP headers and status codes, result types and caching hints, `InputRequired`, subscriptions, the Apps extension, mirrored tool parameters, client-era selection, client input handlers, and custom client transport changes.

The guide leaves two release checks unchecked: official Inspector coverage and Tier 1 client interoperability. Both require real external clients.

No code changes: this PR documents the breaking protocol migration.

### Spec

- [2026-07-28 specification](https://modelcontextprotocol.io/specification/2026-07-28)

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,0 +1,218 @@
+# Upgrade Guide
+
+## Upgrading To 1.x From 0.x
+
+Laravel MCP 1.x implements MCP revision [`2026-07-28`](https://modelcontextprotocol.io/specification/2026-07-28). Existing server registrations, server classes, tools, prompts, resources, authentication middleware, and fluent server tests do not require structural changes.
+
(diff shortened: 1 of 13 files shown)


pushpak1300:
Superseded. The stack is being reordered so every server change lands before the client changes, and the layers are split smaller. Reopened as a new PR at its new position in the stack.

PR #302, opened 2026-08-10
Add the subscriptions listen stream

pushpak1300:
`2026-07-28` removes `resources/subscribe` and the standalone GET stream. A client that wants change notifications opens one long-lived request instead and says exactly what it wants on it.

```json
{
  "method": "subscriptions/listen",
  "params": {
    "notifications": {
      "toolsListChanged": true,
      "resourceSubscriptions": ["file:///project/config.json"]
    }
  }
}
```

The server acknowledges the subset it will honor, then only sends those:

```json
{
  "method": "notifications/subscriptions/acknowledged",
  "params": {
    "_meta": { "io.modelcontextprotocol/subscriptionId": 1 },
    "notifications": { "toolsListChanged": true, "resourceSubscriptions": [...] }
  }
}
```

A server supplies what flows on the stream by overriding one method:

```php
protected function subscriptions(Subscription $subscription): iterable
{
    yield Response::notification('notifications/tools/list_changed');
}
```

Anything yielded that the client did not subscribe to is dropped, including a `resources/updated` for a URI outside `resourceSubscriptions`. Every message on the stream carries the subscription id, and the stream ends with the empty `subscriptions/listen` result that signals a graceful close rather than a dropped connection.

The package never implemented `resources/subscribe` or the GET stream, so nothing had to be removed here and this PR is additive.

Request-scoped notifications like `notifications/progress` keep flowing on their originating response stream, which is where they already were. The `subscriptions()` hook is a generator the server owns, so how it learns about changes, polling, a queue, a broker, is left to the application rather than assumed here. Tests cover acknowledgment ordering and content, the id stamping, the filtering in both directions, and the graceful close.

### Spec

- [Subscriptions](https://modelcontextprotocol.io/specification/2026-07-28/basic/patterns/subscriptions)

### DIFF
--- a/src/Enums/MetaKey.php
+++ b/src/Enums/MetaKey.php
@@ -9,4 +9,5 @@ enum MetaKey: string
     case PROTOCOL_VERSION = 'io.modelcontextprotocol/protocolVersion';
     case CLIENT_CAPABILITIES = 'io.modelcontextprotocol/clientCapabilities';
     case SERVER_INFO = 'io.modelcontextprotocol/serverInfo';
+    case SUBSCRIPTION_ID = 'io.modelcontextprotocol/subscriptionId';
 }
--- a/src/Server.php
+++ b/src/Server.php
@@ -25,6 +25,7 @@ use Laravel\Mcp\Server\Methods\CompletionComplete;
 use Laravel\Mcp\Server\Methods\Concerns\ResolvesResources;
 use Laravel\Mcp\Server\Methods\Discover;
 use Laravel\Mcp\Server\Methods\GetPrompt;
+use Laravel\Mcp\Server\Methods\Listen;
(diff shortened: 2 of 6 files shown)


pushpak1300:
Dropped from the migration: `subscriptions/listen` is optional, not required.

The resources capability is explicit that both features are optional — "Servers **may** advertise either feature independently, together or neither" — and no MUST anywhere obliges a server to implement `subscriptions/listen`. The obligation to push is gated on declaring `listChanged`, which this package defaults to `false`.

Laravel MCP never implemented `resources/subscribe` or the standalone GET stream, so omitting this is not a regression and leaves no compliance gap. Keeping the migration to what the spec actually requires rather than adding an optional feature alongside it. `UPGRADE.md` in #305 now records that the removed methods were never supported and that neither capability is advertised.

PR #310, opened 2026-08-11
Mirror annotated tool parameters into request headers

pushpak1300:
Network intermediaries — load balancers, proxies, WAFs — often need to route on one tool argument, but they cannot afford to parse a JSON-RPC body to find it. `2026-07-28` lets a server nominate arguments to be mirrored into `Mcp-Param-*` request headers so those hops can read them cheaply.

A tool names the arguments it wants mirrored:

```php
class ExecuteSql extends Tool
{
    protected array $parameterHeaders = ['region' => 'Region'];
}
```

The advertised schema then carries the annotation, which is what tells a client to send the header:

```json
"region": { "type": "string", "x-mcp-header": "Region" }
```

and a call with `"region": "us-west1"` is expected to arrive with `Mcp-Param-Region: us-west1`.

### Validating the annotation

`Support\ToolHeaders` enforces the spec's constraints on every advertised schema. A name must be a non-empty RFC 9110 field-name token, must be case-insensitively unique across the schema, must sit on a `string`, `integer`, or `boolean` — never `number` — and must be statically reachable from the schema root, so an annotation buried under `items` or a composition keyword is rejected. Integers outside the IEEE754 safe range cannot be mirrored.

The spec makes rejecting a malformed annotation the *client's* job, and a client that rejects one drops the whole tool from `tools/list`. Validating server-side instead means a bad annotation surfaces as an exception while building the schema, rather than as a tool that silently vanishes for conforming clients.

### Enforcing it on call

Over HTTP, `tools/call` now checks that each annotated argument present in the body arrived with a matching header, failing with `-32020` when one is missing or disagrees with the body. The check is routed from `CallTool` through `ServerContext` because the header middleware cannot reach the tool's schema. Other transports ignore the annotations entirely, which the spec permits.

Sensitive values should not be annotated — header values are visible to every hop in between.

### Spec

- [Tools: `x-mcp-header`](https://modelcontextprotocol.io/specification/2026-07-28/server/tools#x-mcp-header)
- [Custom headers from tool parameters](https://modelcontextprotocol.io/specification/2026-07-28/basic/transports/streamable-http#custom-headers-from-tool-parameters)

Split out of #301, which now covers caching hints only. The client half that mirrors these arguments into request headers is #300, higher in this stack.

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -36,6 +36,7 @@ use Laravel\Mcp\Server\ServerContext;
 use Laravel\Mcp\Server\Testing\PendingTestResponse;
 use Laravel\Mcp\Server\Testing\TestResponse;
 use Laravel\Mcp\Server\Tool;
+use Laravel\Mcp\Server\Transport\HttpTransport;
 use Laravel\Mcp\Transport\JsonRpcNotification;
 use Laravel\Mcp\Transport\JsonRpcRequest;
 use Laravel\Mcp\Transport\JsonRpcResponse;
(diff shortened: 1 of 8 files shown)


pushpak1300:
Superseded: the server-side half of `x-mcp-header` (`Tool::$parameterHeaders` plus the `Mcp-Param-*` validation) is optional per the spec — "the use of `x-mcp-header` is optional for servers" — and is a niche most Laravel apps will never need, so it is dropped rather than shipped as new public API in the migration.

Supporting the annotation *is* required of clients, so the shared `Support\ToolHeaders` contract moved up into #300 alongside its only consumer. Nothing is lost from the migration.

PR #321, opened 2026-08-13
Add rate limiting to the OAuth client registration endpoint

pushpak1300:
This PR adds rate limiting to the OAuth dynamic client registration endpoint, following the same shape Fortify uses:

```php
// config/mcp.php
'limiters' => [
    'register' => 'mcp-register',
],
```

The default is 10 per minute per IP and gets registered in the service provider, so it applies without anyone needing to publish the config first. There are three ways to change it: define your own `mcp-register` limiter, point the config at a different limiter name, or set it to `null` to turn it off.

```php
RateLimiter::for('mcp-register', fn ($request) => Limit::perMinute(60)->by($request->ip()));
```

Registration payloads also now cap `redirect_uris` at 5 entries of 2048 characters each, which every real client fits inside comfortably.

### DIFF
--- a/config/mcp.php
+++ b/config/mcp.php
@@ -53,6 +53,21 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Rate Limiting
+    |--------------------------------------------------------------------------
+    |
+    | By default, the OAuth dynamic client registration endpoint is rate limited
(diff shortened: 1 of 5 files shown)


taylorotwell:
Honestly this will end up blocking so many legit users we probably can't merge.

PR #322, opened 2026-08-14
Expose OAuth challenges on authenticated MCP routes

### LINKED ISSUE #278: AddWwwAuthenticateHeader never runs when the MCP route is protected by auth middleware

Gujiassh:
Fixes #278

## Summary

- Register `AddWwwAuthenticateHeader` in Laravel's HTTP middleware priority list.
- Ensure the middleware wraps authentication so OAuth-protected MCP routes can decorate authentication-generated `401` responses.
- Add a feature regression test covering an unauthenticated request to a protected MCP route.

## Verification

- `vendor/bin/pest --no-coverage` (1028 tests, 2333 assertions)
- `vendor/bin/phpstan --no-progress`
- `vendor/bin/pint --test`
- `vendor/bin/rector --dry-run`

Without the provider change, the new regression test receives a `401` without `WWW-Authenticate`. With it, the response includes the RFC 9728 `resource_metadata` challenge.


pushpak1300:
can you target it to 0.x ?

pushpak1300:
## 🤖 AI Review 

### Problem assessment

The bug (#278) is real and hits the main production path. `Mcp::web(...)->middleware('auth:api')` is the canonical way to deploy an OAuth-protected HTTP MCP server, and the `WWW-Authenticate` challenge is how a spec-compliant MCP client discovers the authorization server (RFC 9728 / MCP auth spec discovery flow). Because Laravel's kernel sorts `Authenticate` (via the `AuthenticatesRequests` contract in `$middlewarePriority`) to the outermost layer, `AddWwwAuthenticateHeader` ends up *inside* auth and never sees the 401 — so the middleware silently no-ops in exactly the configuration it exists for. No error, no log line; only challenge-dependent clients fail.

### Solution assessment

The fix is the right mechanism at the right size: `prependToMiddlewarePriority(AddWwwAuthenticateHeader::class)` makes the sorter place the decorator outside `Authenticate`, and since `Illuminate\Routing\Pipeline` renders the `AuthenticationException` into a 401 inside the pipeline, the now-outermost middleware sees and decorates it. Prepending to the global priority list is inert for non-MCP routes since the middleware is only attached to MCP routes. The regression test covers the previously broken path end to end.

### Suggested improvements (non-blocking)

1. **Loosen the `instanceof \Illuminate\Foundation\Http\Kernel` guard.** An app with a custom kernel that doesn't extend Foundation's silently keeps the bug — the same silent-failure shape this PR fixes. `method_exists($kernel, 'prependToMiddlewarePriority')` covers custom kernels for free:
   ```php
   if (method_exists($kernel, 'prependToMiddlewarePriority')) {
   ```

2. **Avoid clobbering an existing challenge.** Pre-existing behavior, but now that the middleware actually runs on auth 401s it matters: if a guard (e.g. Passport) already emitted its own `WWW-Authenticate`, the middleware overwrites it. A one-line guard in `AddWwwAuthenticateHeader::handle()`:
   ```php
   if ($response->headers->has('WWW-Authenticate')) {
       return $response;
   }
   ```

3. **Test with the `auth` alias instead of `Authenticate::class`.** Real apps write `->middleware('auth:api')` — the configuration reported in #278. `SortedMiddleware` resolves aliases before sorting so both pass, but the alias form is the stronger regression guard.

**Bottom line:** real bug on the main production path, minimal idiomatic root-cause fix. Item 1 is the only change worth pushing for before merge.

pushpak1300:
Thanks

PR #330, opened 2026-08-19
fix: reject non-word URI template placeholders

### LINKED ISSUE #329: Resource template placeholders that are not word characters (e.g. {+path}) silently never match
# Bug Report — laravel/mcp

## Laravel Package Version
v1.0.0-beta.1 (also present on `main` @ a0033a0)

## Laravel Version
13.26.0

## PHP Version
8.5.0

## Database Driver & Version
N/A

---

## Description

A resource template whose placeholder contains anything other than letters, numbers, or
underscores is accepted without complaint, advertised to clients in
`resources/templates/list`, and then **silently fails to match any URI**. No exception,
no warning, no log entry.

`UriTemplate` validates templates with one pattern and parses them with a narrower one:

- Validation (`URI_TEMPLATE_PATTERN`, `src/Support/UriTemplate.php:21`) uses `{[^{}]+}`,
  which accepts `{customer-id}`, `{+path}`, `{user.id}`.
- Parsing (`extractVariableNames()`, `src/Support/UriTemplate.php:91`) uses `\{(\w+)}`,
  which matches none of them.

Because the gate is wider than the parser, those placeholders pass validation and are then
treated as **literal text**: `compileRegex()` finds no variable segment and runs the braces
through `preg_quote()`. The compiled pattern for `file://projects/{+path}` ends up as
`#^file://projects/\{\+path\}$#`, so it only ever matches the literal brace string.

This is the only malformed-template case that fails silently — missing scheme, absent
placeholder, unclosed brace, over-length, and too-many-expressions all throw
`InvalidArgumentException`.

Worth noting the spec angle: the MCP resources specification defines `uriTemplate` in terms
of [RFC 6570](https://datatracker.ietf.org/doc/html/rfc6570), under which both `{+path}`
(reserved expansion) and dotted varnames are valid. This package supports only the `\w+`
subset, so `file:///{+path}` — a natural way to express "match the rest of the path" — is
currently non-functional rather than merely unsupported.

### Why it is hard to diagnose

The failure surfaces as a misleading error. A client that reads a URI matching the intended
template receives:

```
-32602  Resource [acme://customers/451/orders] not found.
```

The message points at the URI, which is correct, rather than at the placeholder, which is
not. The resource is registered, listed, and looks healthy; `handle()` simply never runs.

---

### Minimal

```bash
composer install

php -r 'require "vendor/autoload.php";
$t = new Laravel\Mcp\Support\UriTemplate("file://projects/{+path}");
var_dump($t->variableNames());
var_dump($t->match("file://projects/src/Main.php"));'
```

Actual output — constructed without error:

```
array(0) { }   // variableNames() — the placeholder was never parsed
NULL           // match() — fails against a URI that should match
```

Expected: either an exception at construction, or `variableNames() === ["path"]` and
`match()` returning the captured segment.

### End to end

1. Declare a templated resource:

```php
class ProjectFileResource extends Resource implements HasUriTemplate
{
    public function uriTemplate(): UriTemplate
    {
        return new UriTemplate('file://projects/{+path}');
    }

    {
        return Response::text('path is '.$request->get('path'));
    }
}
```

2. Register it on a server and boot. No error is raised.

3. `resources/templates/list` advertises it to the client:

```json
{ "name": "project-file-resource", "uriTemplate": "file://projects/{+path}" }
```

4. `resources/read` with `{"uri": "file://projects/src/Main.php"}` returns:

```json
{ "error": { "code": -32602,
             "message": "Resource [file://projects/src/Main.php] not found." } }
```

The same happens with a hyphen (`{customer-id}`), which is the likelier accident given how
common hyphens are in Laravel URLs.

---

## Possible directions

Two options, and I did not want to presume which you would prefer:

1. **Reject unsupported placeholders** with a clear `InvalidArgumentException` naming the
   offending placeholder. Small and contained. Trade-off: because `uriTemplate()` is called
   while building the listing, one bad template makes the whole `resources/templates/list`
   response fail rather than just the offending resource.

2. **Implement the RFC 6570 forms** the spec references (`{+var}` at minimum). Spec-correct
   and non-breaking, but a substantially larger change.

I have option 1 implemented with tests and happy to open a PR, or to attempt option 2 if
that is the direction you would rather take. Just let me know which you prefer.

Gujiassh:
## Summary

`UriTemplate` previously validated placeholders with `{[^{}]+}` but only extracted `{\w+}` names. Templates such as `file://projects/{+path}` or `file://customers/{customer-id}` were therefore accepted, advertised to clients, and then treated as literal brace text — so `match()` silently never succeeded.

This change rejects unsupported placeholders at construction time with a clear `InvalidArgumentException`, matching the package's existing fail-fast behavior for other malformed templates.

## Test plan

- [x] `./vendor/bin/pest tests/Unit/Support/UriTemplateTest.php` (22 passed)
- [x] New cases cover `{+path}`, `{customer-id}`, `{user.id}` rejection and plain `{path}` / `{customer_id}` acceptance

Fixes #329

### DIFF
--- a/src/Support/UriTemplate.php
+++ b/src/Support/UriTemplate.php
@@ -82,7 +82,6 @@ class UriTemplate implements Stringable
         );
     }
 
-
     /**
      * Reject placeholders the parser cannot extract.
      *
--- a/src/Support/UriTemplate.php
+++ b/src/Support/UriTemplate.php
@@ -33,6 +33,8 @@ class UriTemplate implements Stringable
             throw new InvalidArgumentException('Invalid URI template: must be a valid URI template with at least one placeholder.');
         }
 
+        $this->assertWordCharacterPlaceholders($template);
+
         $this->variableNames = $this->extractVariableNames($template);
     }
 
@@ -78,6 +80,33 @@ class UriTemplate implements Stringable
             InvalidArgumentException::class,
             sprintf('%s exceeds the maximum length of %d characters (received %d)', $context, $max, Str::length($str))
         );
+    }
+
+
+    /**
+     * Reject placeholders the parser cannot extract.
+     *
+     * Validation previously accepted any `{...}` expression, but
+     * extractVariableNames() / compileRegex() only understand `{word}`
+     * placeholders. Non-word names such as `{customer-id}` or `{+path}`
+     * were therefore advertised and then treated as literal text.
+     */
+    private function assertWordCharacterPlaceholders(string $template): void
+    {
+        if (! preg_match_all('/\{([^{}]+)}/', $template, $matches)) {
+            return;
+        }
+
+        foreach ($matches[1] as $expression) {
+            throw_unless(
+                preg_match('/^\w+$/', $expression) === 1,
+                InvalidArgumentException::class,
+                sprintf(
+                    'Invalid URI template placeholder {%s}: only word-character variable names are supported.',
+                    $expression
+                )
+            );
+        }
     }
 
     /**
--- a/tests/Unit/Support/UriTemplateTest.php
+++ b/tests/Unit/Support/UriTemplateTest.php
@@ -133,3 +133,17 @@ it('casts to string', function (): void {
 
     expect((string) $template)->toBe('file://users/{id}');
 });
+
+it('rejects placeholders that are not word characters', function (): void {
+    expect(fn (): UriTemplate => new UriTemplate('file://projects/{+path}'))
+        ->toThrow(InvalidArgumentException::class, 'only word-character variable names are supported')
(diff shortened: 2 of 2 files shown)


pushpak1300:

We’ve found that AI-generated code often doesn’t align well with Laravel’s conventions, architectural decisions, and the specific context of what we’re trying to accomplish with our framework and packages. Quality contributions require thoughtful human insight into the codebase.

If you’re interested in contributing to Laravel, I’d encourage you to familiarize yourself with the existing codebase, engage with the community, and submit PRs that reflect your own understanding and careful consideration of the problem you’re solving.

PR #331, opened 2026-08-23
Validate params shape on JSON-RPC notifications

shoemoney:
A notification whose `params` member isn't an object throws a TypeError out of the constructor, so the server answers with an internal error instead of a protocol one.

`JsonRpcRequest::from()` has guarded this since #273. Notifications now get the same check and come back as a -32602 invalid params error.

Tests cover the scalar, null, and list cases.


taylorotwell:

PR #335, opened 2026-08-25
Argument attribute

gdebrauwer:
In MCP tools, it happens a lot that you accept one or more model ids. In the tool, you then have to manually convert those to Eloquent models before you can use them
```php
class ChangeTicketStatusTool extends Tool
{
	public function handle(Request $request)
	{
		$ticket = Ticket::find($request->get('ticket_id');
		$status = TicketStatus::find($request->get('ticket_status_id'));

		if ($ticket === null) {
            return Response::error('Ticket not found.');
		}

		if ($status === null) {
            return Response::error('Ticket status not found.');
		}

		// do something ...
	}
}
```

I think it would be nice if there was some model binding similar to route model binding. This PR makes that possible, so you dont have to do all that boilerplate yourself:

```php
class ChangeTicketStatusTool extends Tool
{
	public function handle(
		#[Argument('ticket_id')] Ticket $ticket,
		#[Argument('ticket_status_id')] TicketStatus $status
	) {
		// do something ...
	}
}
```

### DIFF
--- a/src/Attributes/Argument.php
+++ b/src/Attributes/Argument.php
@@ -13,14 +13,21 @@ use Illuminate\Validation\ValidationException;
 use Laravel\Mcp\Request;
 use ReflectionNamedType;
 use ReflectionParameter;
+use RuntimeException;
 
 #[Attribute(Attribute::TARGET_PARAMETER)]
 class Argument implements ContextualAttribute
 {
     public function __construct(public ?string $name = null) {}
 
-    public static function resolve(self $attribute, Container $container, ReflectionParameter $parameter): mixed
+    public static function resolve(self $attribute, Container $container, ?ReflectionParameter $parameter = null): mixed
     {
+        $parameter ??= self::resolveReflectionParameter();
+
+        if (! $parameter instanceof ReflectionParameter) {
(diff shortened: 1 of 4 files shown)


taylorotwell:
Why even need the Argument attribute? Why not just support model binding as normal without any additional attributes?

gdebrauwer:
> Why even need the Argument attribute? Why not just support model binding as normal without any additional attributes?

Currently, in my tools, the arguments are suffixed with `_id`, which differs from the model binding in routes where that is not the case. I just assume that making those argument names very explicit is better for use by the AI, but maybe that is not the case? 🤔  I'm not sure what other people do.

So you would support something like this?
```php
class UpdateTicketDescriptionTool extends Tool
{
	public function handle(Ticket $ticket, string $description)
	{
		// do something ...
	}

	public function schema(JsonSchema $schema): array
    {
        return [
            'ticket' => $schema->integer()
                ->description('The numeric ID of the ticket to delete.'),
			'description' => $schema->string()
				->description('The new ticket description'),
        ];
    }
}
```

taylorotwell:
Oh, gotcha. Yeah makes sense now.

taylorotwell:
Going to skip this one for now but may revisit in the future!

PR #336, opened 2026-09-03
Support nested request input with Laravel-style dot notation

ethanbray:
MCP requests do not support dot notation or wildcards like HTTP requests do in the Laravel core. This recently tripped us up when developing our own MCP. Adding `data_get` to the `Request::data()` method will bring the MCP request functionality inline with HTTP requests.

This is technically a breaking change as current consumers of this package could be using dots in key names e.g.

```php
 public function schema(JsonSchema $schema): array
  {
      return [
          'limit.items' => $schema->integer()
              ->description('Maximum number of items to return')
              ->min(1),
      ];
  }
```

With this change, `data_get` will look for `$this->arguments['limit']['items']` and not fallback to `$this->arguments['limit.items']`.

This is consistent with how the HTTP requests work I believe. Personally I think it's a valid tradeoff considering this project is pre-1.0 and support for dot notation and wildcards is quite helpful.

Let me know if there are any changes I can make to increase the chances of this being accepted.

### DIFF
--- a/src/Request.php
+++ b/src/Request.php
@@ -49,11 +49,7 @@ class Request implements Arrayable
 
     protected function data(mixed $key = null, mixed $default = null): mixed
     {
-        if (is_null($key)) {
-            return $this->arguments;
-        }
-
-        return $this->arguments[$key] ?? $default;
+        return data_get($this->arguments, $key, $default);
     }
 
     public function get(string $key, mixed $default = null): mixed
--- a/tests/Unit/RequestTest.php
+++ b/tests/Unit/RequestTest.php
@@ -69,6 +69,18 @@ it('retrieves nested input using dot notation', function (): void {
         ->and($request->filled('user.country'))->toBeFalse();
 });
 
+it('does not resolve literal dotted keys', function (): void {
+    $request = new Request([
+        'limit.items' => 5,
+        'limit' => ['items' => 10],
+    ]);
+
+    expect($request->get('limit.items'))->toBe(10)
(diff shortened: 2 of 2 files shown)


pushpak1300:
Thanks for your pull request!

Unfortunately, I'm going to delay merging this code for now. To preserve our ability to adequately maintain this package, we need to be very careful regarding the amount of code we include.


If applicable, please consider releasing your code as an additional package so that the community can still take advantage of your contributions!

ethanbray:
@pushpak1300 Can I have some more context on that decision please? While I appreciate you can't bundle every single requested feature into a library, this is literally a one line change that brings the functionality of an MCP request more inline with a Laravel HTTP request. The amount of code being added compared to the benefit it brings is miniscule. I would have to imagine that a developer coming from a standard Laravel application would expect this feature to exist here.

All I'm looking for is something other than a canned response to explain _why_ this does not make sense as part of this package, because the amount of code to maintain is so small that the only reason to reject it would be that this feature is not appropriate for this package?

pushpak1300:
Thanks for pushing back @ethanbray. My initial reading was influenced by the description which focuses on dotted property names in schemas. This isn’t a direction we want: tool argument names should be plain identifiers and nested data should be modelled as nested objects. A change framed as “support limit.items style keys” wouldn’t be merged.

After your comment, I considered it further and the implementation itself aligns with what we want.

Reopening
can you add the following -
- Add a test asserting a literal limit.items key is no longer resolvable to ensure the behaviour is deliberately locked in.

Appreciate the contribution and the nudge.

PR #339, opened 2026-09-11
Updated HasStructuredContent so structured content always gets returned correctly.

Smoseby:
Currently when a indexed array is passed as a structured response, like this for example:

```php
$locations = Location::select('id', 'name')->get();

return $locations->isEmpty() ? Response::text('No results') : Response::structured($locations->toArray());
```

It gets returned in the response body as a array, this works in the MCP inspector since they ran into issues with this and loosened their specs. But as its invalid content according to the MCP spec, applications like Claude throw a "Tool execution failed" error.

I think this is the simplest solution, but id love to hear your take on this.

### DIFF
--- a/src/Server/Concerns/HasStructuredContent.php
+++ b/src/Server/Concerns/HasStructuredContent.php
@@ -31,6 +31,6 @@ trait HasStructuredContent
             return $baseArray;
         }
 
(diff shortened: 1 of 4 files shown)


mikemeijer:
@Smoseby This change has changed my life forever, thank you so much!

PR #340, opened 2026-09-11
Revert "Updated HasStructuredContent so structured content always gets returned correctly."

pushpak1300:
Reverts laravel/mcp#339


pushpak1300:
This breaks on 2026-07-28 where the spec says structuredContent "can be any JSON value (object, array, string, number, boolean, or null)" and ships a `list_users` example returning an array — the cast turns that legal array into `{"0":{...},"1":{...}}`, which matches no `outputSchema` and no longer matches the text block `Response::structured()` encodes at `src/Response.php:97`.

PR #347, opened 2026-09-19
Add WebMCP support for in-page browser agents

pushpak1300:
A remote MCP server needs a full OAuth flow before an agent calls a single tool. The visitor must also add the server to a client config by hand, so your Laravel app stays invisible to the agent that already runs in their browser.

## The API

Turn it on in `.env`:

```
MCP_WEB_MCP_ENABLED=true
```

Add one line to your layout:

```blade
@webMcp
```

That is the whole setup. Every server you route with `Mcp::web()` now exposes its tools, resources, and prompts to in-page agents. Your primitive classes do not change, and `$request->user()` returns the signed in visitor, because the call carries the session cookie.

Hold a primitive back with the `shouldRegister` hook it already has:

```php
class DeleteAccountTool extends Tool
{
    public function shouldRegister(): bool
    {
        return ! Mcp::viaWebMcp();
    }
}
```

This works the same way on a `Resource` or a `Prompt`. It also applies when the primitive runs, not only when it is listed.

Set the middleware for the browser surface in `config/mcp.php`. Add `auth` when every tool needs a user:

```php
'web_mcp' => [
    'enabled' => env('MCP_WEB_MCP_ENABLED', false),
    'middleware' => ['web'],
],
```

Your main MCP route keeps its own middleware, so remote clients still use OAuth.


buismaarten:
In my opinion it's better to publish the JavaScript file to the public folder instead of using a Laravel route. Laravel route has some overhead over static files and isn't worth it.

